PR #5, opened 2022-11-17
Deploy proof of concept

o5v3:
This will render an array of sample images to the website. Some of them using a card component (which is a wrapper for the img html tag for now) we can extend, modify and reuse for our project. The elysium delivery logo uses a static file path, and when the site is deployed to Netlify it *should* change automatically to a Cloudinary url.


index.html
```diff
@@ -47,6 +47,11 @@ <h1>The Regis Altare Birthday Project 2023</h1>
         >
       </footer>
     </section>
+    <section id="card-list">
+      <!--Test image serving-->
+      <img class="card-image" src="images/elysium_delivery_services.jpg">
+      <tcg-card></tcg-card>
+    </section>
```
miffens: move this above the footer
side note: the existing main/ section tags organization aren't quite right now that I look at it again.. I think it should be
```
<main>
  <article>
     <h1><h1>
       <footer></footer>
  <article>
<main>
```
No need to worry about all that in your PR, I can do it in a separate PR

index.html
```diff
@@ -47,6 +47,11 @@ <h1>The Regis Altare Birthday Project 2023</h1>
         >
       </footer>
     </section>
+    <section id="card-list">
+      <!--Test image serving-->
+      <img class="card-image" src="images/elysium_delivery_services.jpg">
+      <tcg-card></tcg-card>
```
miffens: Does this `<tcg-card></tcg-card>` line do anything? Delete if not?

index.html
```diff
@@ -47,6 +47,11 @@ <h1>The Regis Altare Birthday Project 2023</h1>
         >
       </footer>
     </section>
+    <section id="card-list">
+      <!--Test image serving-->
+      <img class="card-image" src="images/elysium_delivery_services.jpg">
```
miffens: I checked the deploy preview, and it indeed converted this into a Cloudinary URL :tada: 
So my understanding is if this image were hypothetically not on the home page but on some other page, there would be no network bandwidth consumed when the home page loads in?
In contrast, if we did not use the cloudinary-netlify plugin, this image would use network bandwidth - even if it's not immediately needed for the current page - because it's a static image hardcoded into our website's files.
Is that right?

js/main.js
```diff
@@ -0,0 +1,45 @@
+//Happy Birthday Leader! 🎇💙
+
+const card_list = document.getElementById("card-list");
+const CLOUD_NAME = "dazcxdgiy";
+const CLOUDINARY_URL = `https://res.cloudinary.com/${CLOUD_NAME}/image/upload/`;
```
miffens, line 5: I'm surprised there's no API secret stuff in this diff. Thinking some more, I think that makes sense since we're only "reading" from Cloudinary. API keys (and keeping that secret) only come into play if we're concerned about making upload/ post calls. So there's no security risk of our cloud name being public. Is that right?

js/main.js
```diff
@@ -0,0 +1,45 @@
...
+      this.hasAttribute("card-id")
+        ? `cld-sample-${this.getAttribute("card-id")}`
+        : "sample.jpg"
+    }`;
+  }
+}
+
+//Renders some sample images from Cloudinary's sample folder using our Card component.
+function renderImages() {
+  for (let i = 2; i <= 5; i++) {
```
miffens, line 32: Is there no way to get all the images (or id's) from a given folder via the cloudinary API? Otherwise I'm not sure if there's any effortless way we can keep this loop "synced" when new cards get added. We would have to manually update the code each time.

js/main.js
```diff
@@ -0,0 +1,45 @@
...
+    image.src = this.getImageURL();
+    image.classList.add("card-image");
+    this.appendChild(image);
+  }
+
+  //This returns an url of the form:
+  //`https://res.cloudinary.com/${CLOUD_NAME}/image/upload/cld-sample-${CARD_ID}`
+  //if "card-id" isn't supplied it returns a F L O W E R 🌸 sample image url.
+  getImageURL() {
+    return `${CLOUDINARY_URL}${
```
miffens, line 22: You can do this in a different PR, but is there a way we can also get a given image's metadata? I'm thinking about how we can fetch a card's text description, rarity, etc.

index.html
o5v3: Moving it now o7

index.html
o5v3: It's there to show we can use the component the same way as any other html tag (it's the red flower image), but we can delete it.

index.html
o5v3: Even if we didn't use the plugin, the image wouldn't be requested if we don't use it in that page. The benefit the plugin provides us is that instead of relying on Netlify bandwidth for our entire project, we're splitting it into Netlify bandwidth for HTML, CSS, JS and other files; and Cloudinary bandwidth for images. (Also, Cloudinary has image transformations which could be useful for us, but that's another discussion)

js/main.js
o5v3: Correct, the cloud name is public and can be seen on all image urls, so it's safe to show client side. The API keys are only used at build time to upload the images, so they're set as environment variables inside Netlify.

js/main.js
o5v3: There is a way to do it, but it requires setting tags for the images [(docs)](https://cloudinary.com/documentation/advanced_url_delivery_options#client_side_asset_lists). We would need to either do it manually or write a small script to do it. There's a way to get a list directly from folders or even from filenames [(docs)](https://cloudinary.com/documentation/search_api) , but it requires the API key and secret so it would need either showing them client side (a big no-no), or using Netlify Functions to act as a backend. We can also go the CSV route and just plop the CSV in the files and access it through javascript. This needs extended discussion.

js/main.js
o5v3: Yes, the same methods that give lists of images return their metadata. Here's an example, where the logo and the flower image have the "logo" tag and some metadata: [(example)](https://res.cloudinary.com/dazcxdgiy/image/list/logo.json).

miffens: How did Cloudinary get the alt text etc? Did you set that in the Cloudinary UI?

js/main.js
miffens: Good to know about the multiple options!

index.html
miffens: ah, makes sense! Since you're the sole dev building on top of these changes and you know you'll remove it eventually, you can keep it for now

miffens (approved):

js/main.js
o5v3: Yep. You can open the media library, open an image, go to the metadata tab and set everything there.

PR #6, opened 2022-11-24
Add nav w/ basic "binder tab" look, "birthday msg board" page w/ test video and kudoboard link

miffens:


pages/message-board.html
```diff
@@ -0,0 +1,70 @@
...
+          >Go to message wall</a
+        >
+        <video autoplay muted loop controls>
+          <source
+            src="https://res.cloudinary.com/dazcxdgiy/video/upload/v1669320701/kudoboard_qbu1gi.mp4"
+            type="video/mp4"
+          />
+          Your browser does not support the video tag.
+        </video>
+      </section>
```
miffens: rest of file is copy paste from index.html. Here are the main changes. Notes:
 - The "active" class hardcoded on the "current" page is what achieves the hover styling only being applied to non-active tabs. You can also dynamically compute this with JS, but I felt our project is simple enough to just hardcode.
 - 'muted' must be set with 'autoplay' for it to work in most browsers

miffens: I made a new pages folder since that's what the repo we're referencing did, but tbh remembering to update the file paths was a bit tedious. If anyone prefers having all the html pages be on the same level as index.html I can change it.

o5v3: I feel that having the pages in a designated folder is better for organization, to have everything in its place instead of cluttering the root, so I'm fine with it. And yeah, we don't have many pages so hardcoding the class should be fine. If it turns out the project needs way more pages for some reason we can always change it to dynamic styling.

o5v3 (approved):

PR #7, opened 2022-11-30
Implement Papa Parse to parse CSV card data file

o5v3:
Adds the Papa Parse library and uses it to read the CSV file located at root. It can also access CSV inside folders by changing the CSV_FILENAME constant and using the Papa.parse() function.
You can check if it works by opening the browser console and printing the value of cards_data.


js/main.js
```diff
@@ -1,8 +1,13 @@
 //Happy Birthday Leader! 🎇💙
+//import Papa from "papaparse"
```
miffens: does this line do anything if it's commented? delete?

miffens (approved):

js/main.js
o5v3: It doesn't do anything. On second thought it confuses more than it helps, so already deleted it. o7

PR #8, opened 2022-12-07
Implement basic gacha pull functionality

o5v3:
This one is relatively big compared to the others, these are the modifications:

- Moved the card component's HTML to its own file to allow for easier access and editing.
- Made an extremely barebones gacha page.
- Implemented functions for shuffling and getting random sets of cards.
- Also implemented a way to allow selective execution based on the current page.
- Allowed cards to be flipped on pages inside the PAGES_WHERE_CARD_HIDDEN constant array. Everywhere else shows the card art right away.
- Keyboard users can also flip cards with the Enter key.

The javascript file is getting big, so I may split it in the next pull request and start using many files to manage JS instead.


js/main.js
```diff
@@ -1,34 +1,67 @@
...
-    return `${CLOUDINARY_URL}${
-      this.hasAttribute("card-id")
-        ? `cld-sample-${this.getAttribute("card-id")}`
-        : "sample.jpg"
-    }`;
+    setupFlipping() {
+      const card_cover = this.getElementsByClassName("card-cover")[0];
+      card_cover.onclick = (event) => this.flipCard();
+      //Necessary for keyboard focus
+      card_cover.setAttribute("tabindex", 0);
```
miffens: If you make the card a `<button>`, you won't need to do this, and it will also tell screenreaders that this is a clickable element. Strip the browser default button styling, and you should be able to style the card "button" as you would a div

The other finesse for screenreader compatibility - to communicate that clicking the first time flips the card, clicking the second onwards opens the detail popup - is probably worth its own focused PR. I could take it while you continue fleshing out the gacha logic - whichever you're more excited to work on

js/main.js
```diff
@@ -38,24 +71,60 @@ function getCSV() {
...
     );
   }
 }
 
-function main() {
-  customElements.define("tcg-card", Card);
-  renderImages();
+async function main() {
+  await defineCardComponent();
   getCSV();
```
miffens: does `getCSV` also need an await?

pages/gacha.html
```diff
@@ -0,0 +1,68 @@
...
+  </nav>
+
+  <section id="gacha-controls">
+    <button id="gacha-button">
+      Pull Gacha
+    </button>
+  </section>
+
+  <section id="card-list">
+  </section>
```
miffens: Move this within the "gacha-controls" section (`id` probably will be renamed) to have a single "main body" instead of 2, leading to this weird border:
Screenshot 2022-12-07 180426

Nesting section tags is allowed for conveying the page's structure ("within the main gacha game section, we have the gacha controls section and the cards")

js/main.js
```diff
@@ -38,24 +71,60 @@ function getCSV() {
...
+    let i = Math.floor(Math.random() * len--);
+    let j = samples[len];
+    samples[len] = samples[i];
+    samples[i] = j;
+  }
+  return samples.slice(0, n);
+}
+
+//Pulls "n" number of cards from the cards_data array and renders them in CARD_LIST.
+function pullGacha(n) {
```
miffens: (nit) rename to something like `pullAndRenderCards`. Just from reading the names, it took me a moment to figure out the difference between `pullCards` and `pullGacha`

index.html
```diff
@@ -31,13 +31,11 @@ <h1>The Regis Altare Birthday Project 2023</h1>
 
       <nav>
         <a class="active" href="index.html">Collection</a>
```
miffens: Soon I think `index.html` will become the amazing welcome page Bon is building. Maybe in this PR or a quick followup we can refactor the current logic to move "collection" off its own page and let index.html be blank for Bon to fill in.

js/main.js
```diff
@@ -1,34 +1,67 @@
 //Happy Birthday Leader! 🎇💙
 
-const card_list = document.getElementById("card-list");
+const CARD_LIST = document.getElementById("card-list");
```
miffens: (nit) maybe something like ALL_CARDS_PAGE or COLLECTIONS_MAIN_CONTENT is a better name for this, else I'm unsure if this is a "card list" for the gacha page or what

js/main.js
```diff
@@ -1,34 +1,67 @@
...
+const GACHA_BUTTON = document.getElementById("gacha-button");
 const CLOUD_NAME = "dazcxdgiy";
 const CLOUDINARY_URL = `https://res.cloudinary.com/${CLOUD_NAME}/image/upload/`;
-const CSV_FILENAME = "Test Card List CSV.csv";
+const CSV_FILENAME = "../Test Card List CSV.csv";
+const CARDS_PER_PULL = 10;
+let pathname = window.location.pathname;
+const CURRENT_PAGE = pathname.slice(pathname.lastIndexOf("/"), pathname.length);
```
miffens, line 10: I didn't expect CURRENT_PAGE to dynamically update when the page changes, but looks like it works 😮

js/main.js
```diff
@@ -1,34 +1,67 @@
...
-  //This returns an url of the form:
-  //`https://res.cloudinary.com/${CLOUD_NAME}/image/upload/cld-sample-${CARD_ID}`
-  //if "card-id" isn't supplied it returns a F L O W E R 🌸 sample image url.
-  getImageURL() {
-    return `${CLOUDINARY_URL}${
-      this.hasAttribute("card-id")
-        ? `cld-sample-${this.getAttribute("card-id")}`
-        : "sample.jpg"
-    }`;
+    setupFlipping() {
```
miffens: (nit) add comments here describing what this function does, like
"set up user interaction for flipping cards"

js/main.js
```diff
@@ -1,34 +1,67 @@
...
+        if (event.key === "Enter") {
+          this.flipCard();
+        }
+      };
+    }
   }
+  customElements.define("tcg-card", Card);
 }
 
 function getCSV() {
```
miffens: (nit) the name `getCSVAndMaybeRenderCollection` would be clearer. That is a mouthful though..

js/main.js
```diff
@@ -38,24 +71,60 @@ function getCSV() {
...
+//Pulls "n" number of cards from the cards_data array and renders them in CARD_LIST.
+function pullGacha(n) {
+  let pulled = pullCards(cards_data, n);
+  renderCards(pulled, CARD_LIST, true);
+}
+
+//Renders a list of cards in the element specified in htmlLocation.
+//If replace is true, overwrites all elements inside htmlLocation.
+//else, adds the cards to the rest of the inner content.
+function renderCards(cards, htmlLocation, replace = false) {
```
miffens, line 107: It was a bit counterintuitive to me that `renderCards` and `defineCardComponent` are intertwined but are far apart in this file. No changes needed for this PR, but when you split out main.js, my idea is to have these two functions in the same file

miffens (approved):
I usually work in a framework, so it was really cool to see how to organize code/ implement using vanilla JS!

js/main.js
o5v3: I'm learning a lot about accessibility with this, thank you! Actually, the change to buttons makes it easier to work with in javascript too, so it hits two birds with one stone.

Personally I'm excited for the gacha part and you are more experienced with accessibility, so I'd be thankful if you could take care of that part. 🙇

js/main.js
o5v3: Weirdly, Papa parse doesn't use promises for async operations, it uses callbacks instead. I'd love to use an await for it though, it'd make execution flow a lot easier to follow.

pages/gacha.html
o5v3: Done. o7

js/main.js
o5v3: Done. o7

index.html
o5v3: Moved to its own page. o7

js/main.js
o5v3: Done. o7

js/main.js
o5v3: It wouldn't work if this was an SPA. One of the only perks of serving static html. 😎

js/main.js
o5v3: Done. Also changed name for greater clarity.

js/main.js
o5v3: Yeah, that's why I'd prefer if Papa used async. I'll try to find a way to make it clearer for next PR.

js/main.js
o5v3: Sure thing. 👍

PR #9, opened 2022-12-08
Welcome screen setup

bonbombs:
Probably will wait until https://github.com/elysiumdelivery/altare-tcg/pull/8 gets merged in and pull those changes into this branch when that happens!

## HTML tweaks

- Rename the existing index.html to collection.html. Let me know if collections should exist in the pages folder now.
- Setting up the new index.html to include the welcome screen elements (loading screen, HBD prompt, main navigation w/ splash art)

## CSS tweaks

- Adding welcome.css to style to index.html welcome screen elements & animations
- Additional styling to page for mobile/narrow viewports
- Moving nav css to nav.css for modularity & reuse in non-welcome screen pages
- Setting main.css as the file containing styles that pertain to all pages

Will likely make adjustments to colors & element styling down the line based on feedback received.

## JS tweaks

- Adding welcome.js includes welcome screen specific functions

## Other

- Adding temp splash page art


index.html
```diff
@@ -19,27 +19,46 @@
...
+      </section>
 
-      <section id="card-list">
-        <!--Test image serving-->
-        <img class="card-image" src="images/elysium_delivery_services.jpg" />
-        <tcg-card></tcg-card>
+      <section id="welcome-splash-2" class="welcome-splash">
+        <img id="altare-img" src="images/AltareSplash.png" />
+        <nav id="welcome-nav">
+          <a href="index.html">Play Gacha</a>
```
miffens: Should 'Play Gacha' be the center card and bigger/ sparklier than the rest? Since that'll be our main attraction

js/welcome.js
```diff
@@ -0,0 +1,19 @@
+const splash_overlay = document.getElementById("splash-overlay");
+const welcome_splash_1 = document.getElementById("welcome-splash-1");
+const welcome_splash_2 = document.getElementById("welcome-splash-2");
```
miffens: (nit) names like `welcome_message` and `welcome_fancy_menu` would be clearer for me to distinguish sections

index.html
```diff
@@ -19,27 +19,46 @@
...
+          <a href="collection.html">Full Deck</a>
+          <a href="pages/messages.html">Artist/Writer Bday Messages</a>
+          <a href="pages/message-board.html">Kudoboard</a>
+          <a href="pages/credits.html">Credits</a>
+        </nav>
       </section>
 
+      <div id="splash-overlay" class="active">
+        <span>Loading</span>
```
miffens: For accessibility, I think it'd be the best experience to automatically narrate changing items on the screen. Adding `aria-live="polite"` will achieve this

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

If you're not confident about addressing all the accessibility comments here, I can take on that work to add them, it wouldn't take me long o7

index.html
```diff
@@ -19,27 +19,46 @@
...
-      <nav>
-        <a class="active" href="index.html">Collection</a>
-        <a href="pages/message-board.html">Birthday Message Board</a>
-      </nav>
+      <section id="welcome-splash-1" class="welcome-splash">
+        <h2>Happy Birthday Altare!</h2>
+        <h3>Let's see what the Altventurers have made for you</h3>
+        <!-- TODO: change this to an svg or img of an arrow -->
+        <button onclick="toggleWelcomeTwo()">arrow</button>
+      </section>
```
miffens: Perhaps "Happy Birthday Altare!..." etc can be in a live narration region, to signal that the loading is complete.

index.html
```diff
@@ -19,27 +19,46 @@
...
 
-      <section id="card-list">
-        <!--Test image serving-->
-        <img class="card-image" src="images/elysium_delivery_services.jpg" />
-        <tcg-card></tcg-card>
+      <section id="welcome-splash-2" class="welcome-splash">
```
miffens: When this appears, perhaps "A fullscreen splash art of Altare leaping greets you with 5 cards for you to select" should be announced
We could use someone who is better at alt text's help here to make the description better haha

I would achieve this by making a live-region div with the visually-hidden class from here: https://www.a11yproject.com/posts/how-to-hide-content/

css/welcome.css
```diff
@@ -0,0 +1,252 @@
...
+    opacity: 1;
+    transform: translateY(0%) !important;
+    pointer-events: auto;
+}
+
+#altare-img {
+    position: absolute;
+    top: 0;
+    left: auto;
+    /* width: 100%; */
```
miffens: delete comment?

index.html
```diff
@@ -19,27 +19,46 @@
...
+        <nav id="welcome-nav">
+          <a href="index.html">Play Gacha</a>
+          <a href="collection.html">Full Deck</a>
+          <a href="pages/messages.html">Artist/Writer Bday Messages</a>
+          <a href="pages/message-board.html">Kudoboard</a>
+          <a href="pages/credits.html">Credits</a>
+        </nav>
       </section>
 
+      <div id="splash-overlay" class="active">
```
miffens: (nit) loading-overlay would be a clearer name for me. All the screens having "splash" made it hard to distinguish them :')

css/main.css
```diff
@@ -1,22 +1,18 @@
...
-  border-top-right-radius: 20px;
+body {
+background: var(--accent-color);
+font-family: 'Inter', sans-serif;
 }
 
-nav a:hover:not(.active) {
-  background: grey;
+h1, h2, h3, h4, h5, h6 {
+font-family: 'Zen Dots', cursive;
```
miffens: would it be worth specifying font fallbacks if Zen Dots doesn't load? I don't know much about fonts

miffens (commented):
Taking an early look since I know the CSS animations are basically done

index.html
bonbombs: I agree. For now, I added a `special-nav-card` class to Play Gacha and it has a slightly different styling to it. We can def change it down the line. Also, switched the CSS positioning of the Play Gacha card with Credits so it's in the center. The order in HTML is the same to keep read order.

css/main.css
bonbombs: `cursive` would be the fallback here. (see https://www.w3schools.com/css/css_font_fallbacks.asp)

Though looking at the cursive fallback it doesn't look great. Changed it to use the `fantasy` fallback font instead lol;;;

index.html
bonbombs: I haven't worked in depth w/ accessibility & ARIA so I may need additional pointers. I made another commit with an attempt. Please let me know if it's incorrect and I'll remove those changes.

css/main.css
```diff
@@ -40,3 +38,19 @@ footer {
...
   width: 1em;
 }
+
+#page-header {
+  color: white;
+  text-shadow: 4px 4px 1px rgb(13 66 163), -3px -3px 1px rgb(148 186 255);
+}
+
+/* screen reader friendly styling */
+.visually-hidden {
```
miffens: This class is meant to be for text that is meant for screenreader users only, not for hiding elements in general. My idea was to only apply the visually-hidden class on the alt text for the fancy menu ("Altare leaps down and there are 5 cards for you to choose from") - that text is not needed for users that can see the screen

But maybe we should wait on setting up visually-hidden alt text until the end when the design is finalized, so I'd be fine with deleting this class and not worrying about getting accessibility perfect until the design is final

index.html
```diff
@@ -19,27 +19,46 @@
...
 
   <body>
     <main>
-      <h1>The Regis Altare Birthday Project 2023</h1>
 
-      <nav>
-        <a class="active" href="index.html">Collection</a>
-        <a href="pages/message-board.html">Birthday Message Board</a>
-      </nav>
+      <section id="welcome-message" class="welcome-splash visually-hidden" role="region" aria-live="polite">
```
miffens: the region role is specified by default for section tags, so you can delete role=region

index.html
```diff
@@ -19,27 +19,46 @@
...
+          <a class="special-nav-card" href="index.html">Play Gacha</a>
+          <a href="collection.html">Full Deck</a>
+          <a href="pages/messages.html">Artist/Writer Bday Messages</a>
+          <a href="pages/message-board.html">Kudoboard</a>
+          <a href="pages/credits.html">Credits</a>
+        </nav>
       </section>
 
+      <div id="loading-overlay" class="active">
+        <span>Loading</span>
```
miffens: 1. Add aria-live = polite on the loading span so that the screenreader reads it aloud, letting the user know not to interact with the page yet
2. Because the text is still in the DOM and just hidden with CSS, I can actually still "see" the "loading" text when using a screenreader
Here is me going through the main fancy menu as a screenreader user. 

https://user-images.githubusercontent.com/47371080/206876218-9bf7ba29-ac52-4ba3-b1a8-5a0d1834db60.mov

It's inaccurate that "happy birthday Altare..." (lol "Al-tar") and "loading" are still being read - they are not visible on the main fancy menu.

According to https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html , using `display: none` (or any of the other things he suggests) will hide it from both visually and from screenreaders.

index.html
```diff
@@ -19,27 +19,46 @@
...
+        <h3>Let's see what the Altventurers have made for you</h3>
+        <!-- TODO: change this to an svg or img of an arrow -->
+        <button onclick="toggleWelcomeTwo()">arrow</button>
+      </section>
 
-      <section id="card-list">
-        <!--Test image serving-->
-        <img class="card-image" src="images/elysium_delivery_services.jpg" />
-        <tcg-card></tcg-card>
+      <section id="welcome-fancy-menu" class="welcome-splash visually-hidden" role="region" aria-live="polite">
```
miffens: Remove `aria-live="polite"` here - it does not do anything because there is no text that is a direct child of it

My idea to provide alt text for the Altare illustration and announce that the page has new content in it would be like:
```
<div aria-live="polite" class="visually-hidden"> Altare leaps out, and there are 5 cards you can choose from </div>
```
added within this section
The primary purpose of this text would be to announce that the page has new content moreso than provide a thorough description of the artwork

But yeah, maybe let's not worry about this for now until the design is more final

index.html
```diff
@@ -19,27 +19,46 @@
...
+        <img id="altare-img" src="images/AltareSplash.png" />
```
miffens: Add `alt="A full screen splash of Altare jumping down from above"` to the `img`
Else it reads as "unlabeled image" which you can hear in the video recording I shared in a diff comment

css/welcome.css
```diff
@@ -0,0 +1,258 @@
+/* welcome page styling goes here */
```
miffens, line 1: idk if it's Netlify's deploy preview being weird since when I shrink my browser your changes look great, but when I open it on my phone it looks like this:
IMG_2914

Can you double check on your end?

miffens (commented):
Remember to resolve merge conflicts as well ^^

css/welcome.css
```diff
@@ -0,0 +1,258 @@
...
+    }
+}
+
+@media only screen and (min-width: 600px) {
+
+    #welcome-nav a:hover, #welcome-nav a:focus-visible {
+        transform: scale(1.1) rotate3d(0, 0, 0, 0deg) !important;
+    }
+    
+    #welcome-nav a:nth-child(1) {
```
miffens, line 239: No changes needed now - FYI the best practice is to have buttons/ links ordered top/ down, left/ right to make it a predictable order for users tabbing through the page. In our context I think it's fine/ cute to break that rule for just the "Play Gacha" card, but I would expect the other cards' tab order to go left > right. Right now it appears to be going inside out

We can address this later when the design is more final

css/welcome.css
bonbombs: hmm odd... I'm opening the page from my mobile brave browser and it seems fine... even after tweaking display settings. (using a Pixel 4a)

what kind of phone are you using?

index.html
bonbombs: Ahh I see! Do you have a link to a guide on how to enable a screen reader? That way I can do some testing on my end while building also

css/welcome.css
miffens: iphone 8. It's probably just me if you don't see it on your phone

index.html
miffens: Mac has VoiceOver built in, which is what I was using. Windows 11 also apparently has a built in screenreader. But if you don't have those, I believe [ChromeVox](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en) is the easiest to set up, but the canonically most popular screenreaders is JAWS (have to restart your computer to get in a 30 minute timed developer mode) or NVDA.

Whichever screenreader you use has specific controls, but the main ones are the application chooser (so you can go to Chrome) and go to next/ previous item on page. Each one will also have subtle differences in reading, but we don't need to get into the weeds on that for this project imo. Some may not have a text box on the screen that displays the text being narrated or a box highlighting the element currently being read out.

lmk if you have more questions!

miffens: PR approval pending the `display: none` stuff o7

bonbombs: I think I'm starting to understand the screen reader accessibility stuff! Being able to see it in action really helped piece things together.
I opted to use `visibility:hidden` over `display:none` since the elements that needed to be marked as hidden had transitions attached to them. I think it did the trick. Did a preliminary test using VoiceOver and it only read each section as they appeared, which is I think what we want. Also added a `visually-hidden` span before the `<nav>` to prompt the user to select a nav card. Actual content of the prompt can be changed later on of course.

index.html
```diff
@@ -19,25 +19,46 @@
...
+      <section id="welcome-message" class="welcome-splash" aria-live="polite">
+        <h2>Happy Birthday Altare!</h2>
+        <h3>Let's see what the Altventurers have made for you</h3>
+        <!-- TODO: change this to an svg or img of an arrow -->
+        <button id="welcome-splash-button" type="button" onclick="toggleWelcomeTwo()">Enter site</button>
       </section>
 
+      <section id="welcome-fancy-menu" class="welcome-splash" aria-live="polite">
+        <img id="altare-img" src="images/AltareSplash.png" alt="A full screen splash of Altare jumping down from above"/>
+        <span class="visually-hidden">Select the following navigation options:</span>
```
miffens, line 52: (nit) The `nav` tag is self explanatory (I hear VO say something like "start of navigation" and "end of navigation") so I don't think this visually hidden text is needed
Oh unless the idea was for this to auto-narrate when the menu loads in? That would make sense

index.html
```diff
@@ -19,25 +19,46 @@
...
-      <nav>
-        <a class="active" href="pages/collection.html">Collection</a>
-        <a href="pages/gacha.html">Gacha</a>
-        <a href="pages/message-board.html">Birthday Message Board</a>
-      </nav>
+      <div id="loading-overlay" class="active">
+        <span aria-live="polite">Loading</span>
+        <span aria-hidden="true">.</span>
+        <span aria-hidden="true">.</span>
+        <span aria-hidden="true">.</span>
```
miffens, line 40: Nice - aria-hidden is correct to use here

js/welcome.js
```diff
@@ -0,0 +1,25 @@
...
+        loading_overlay.classList.remove("active");
+        setTimeout(function () {
+            welcome_message.classList.add("active");
+        }, 100);
+    }, 2000);
+}
+
+function toggleWelcomeTwo () {
+    welcome_message.classList.add("scroll-up");
+    welcome_message.setAttribute('aria-hidden', 'true');
```
miffens: (nit) is aria-hidden needed if you already have visibility: hidden?

miffens (approved):

index.html
miffens: yeah! Happy to hear you got a screen reader set up. I tested it on my end as well and it sounds good. Good call on visibility instead of display.

js/welcome.js
bonbombs: good point! I'll remove that.

index.html
bonbombs: Yes, that was my intention to auto-narrate when the menu loads in.

PR #10, opened 2022-12-12
Open dialog w/ details on clicking card front

miffens:
I'm not a native vanilla JS programmer. Feel free to suggest "better" ways of writing what I did or cleaning it up for me in a future PR!

Clicking on a flipped, front-facing card will pop up an accessible detail dialog (toggles aria-hidden, while the dialog is open, pressing tab does not reach the elements beneath the dialog, esc/ click out to close)


card.html
```diff
@@ -1,8 +1,8 @@
 <div class="card">
-    <button class="card-front">
```
miffens, line 2: prettier automatically did this

css/card.css
```diff
@@ -1,29 +1,85 @@
...
 .card-cover {
-    margin: 3%;
-    border: solid;
-    border-color: dodgerblue;
+  margin: 3%;
+  border: solid;
+  border-color: dodgerblue;
+}
+
+.dialog-title {
```
miffens, line 30: Dialogs and cards are intertwined, but might be worth putting dialog in a separate file or making this file named card-and-dialog

Will decide later after merging in o5v3's upcoming file reorg

css/card.css
```diff
@@ -1,29 +1,85 @@
...
+
+#card-details-dialog-close {
+  grid-area: x;
+}
+
+[role="alertdialog"],
+[role="dialog"] {
+  top: 2rem;
+  left: 50vw; /* move to the middle of the screen (assumes relative parent is the body/viewport) */
+  transform: translateX(-50%); /* move backwards 50% of this element's width */
```
miffens, line 53: copy/pasted these, don't fully understand them :'D

css/main.css
```diff
@@ -40,3 +40,7 @@ footer {
 footer svg {
   width: 1em;
 }
+
+.hidden {
+  display: none;
+}
```
miffens, line 64: Moved this class to global level

package.json
```diff
@@ -1,7 +1,12 @@
 {
   "name": "altare-tcg",
   "version": "0.0.1",
+  "type": "module",
```
miffens: had this as part of attempt to get `import A11yDialog from 'a11y-dialog'` to work. Don't know if it's needed

pages/gacha.html
```diff
@@ -1,68 +1,75 @@
...
-  />
-  <meta name="twitter:card" content="summary" />
-  <meta name="twitter:title" content="Regis Altare Booster Pack Simulator" />
-  <meta
-    name="twitter:description"
-    content="Regis Altare themed booster pack opening game. Open packs and collect all the cards online in this free web app."
-  />
-  <!-- Enables media queries on mobile -->
-  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <head>
```
miffens, line 3: Prettier :woman_shrugging:

js/main.js
```diff
@@ -122,6 +159,10 @@ async function main() {
   if (CURRENT_PAGE == "/gacha.html") {
     GACHA_BUTTON.onclick = (event) => pullAndRenderCards(CARDS_PER_PULL);
   }
+  if (CURRENT_PAGE == "/gacha.html" || CURRENT_PAGE == "/collection.html") {
+    DETAILS_DIALOG_CLOSE.onclick = (event) =>
+      DETAILS_DIALOG.classList.add("hidden");
+  }
```
o5v3, line 183: This part can be made clearer using a switch statement with fall-through instead of ifs. Comes down to preference though.

package.json
o5v3: Only needed to use ES6 module syntax for imports. Not needed right now, but doesn't hurt either.

js/main.js
```diff
@@ -55,12 +59,45 @@ async function defineCardComponent() {
...
+    "details-dialog-text"
+  )[0].insertAdjacentHTML(
+    "beforeend",
+    `
+    <h3>${data["Attack 1 Name"]}</h3>
+    <p>${data["Attack 1 Description"]}</p>
+    <p><i>${data["Attack 1 Extended Lore"]}</i></p>
+    <h3>${data["Attack 2 Name"]}</h3>
+    <p>${data["Attack 2 Description"]}</p>
+    <p><i>${data["Attack 2 Extended Lore"]}</i></p>
```
o5v3, line 114: Should we keep HTML inside the JS component definitions or should we use individual HTML files/templates and import them? Both are valid cases. If everyone involved is fine with manipulating HTML in JS, then no problem here.

Test Card List CSV.csv
```diff
@@ -11,4 +15,4 @@ SecretRare,SecretRare_Rainbowwww_TheGayAgenda,TheGayAgenda,,,,,,,,,,,,,,,,"this,
 UltraRare,UltraRare_GuildHalleyo_this_is_a_test_of_space,this is a test of space,,,,,,,,,,,,,,,,"""item"" # 3",oh yeah,:O,5,GuildHalleyo,e
 UltraRare,UltraRare_OMbretest_Whatiscolour,Whatiscolour,Noble,675,UwU,UwU,def,2,1,dolor,lorem text,Game,yep,2,45,xd,rawr,,,UwU,6,OMbretest,f
 Uncommon,Uncommon_AltareTEST_Slimies,Slimies,,,,,,,,,,,,,,,,is this the last one?,let me check,:3,7,AltareTEST,g
-Uncommon,Uncommon_MagniTEST_Doopdedoo,Doopdedoo,Initiate,7,Zatsu,Music,ghi,1,2,sit,here,,,,,,,,,>:3,8,MagniTEST,h
\ No newline at end of file
+Uncommon,Uncommon_MagniTEST_Doopdedoo,Doopdedoo,Initiate,7,Zatsu,Music,ghi,1,2,sit,here,,,,,,,,,>:3,8,MagniTEST,h
```
o5v3: There's a stray newline here at the end, it's making a new "card" entry and that's why the console is throwing errors.

js/main.js
```diff
@@ -119,8 +174,12 @@ function renderCards(cards, htmlLocation, replace = false) {
 async function main() {
   await defineCardComponent();
   getCSVAndMaybeRenderCollection();
-  if (CURRENT_PAGE == "/gacha.html") {
-    GACHA_BUTTON.onclick = (event) => pullAndRenderCards(CARDS_PER_PULL);
+  switch (CURRENT_PAGE) {
+    case "/gacha.html":
+      GACHA_BUTTON.onclick = (event) => pullAndRenderCards(CARDS_PER_PULL);
```
miffens, line 179: Halp: currently the gacha button doesn't work - GACHA_BUTTON corresponds to the correct element, but the onclick doesn't seem to fire at all (pullAndRenderCards is never reached when I put a breakpoint)

```
      GACHA_BUTTON.onclick = (event) => {
        debugger; // this breakpoint never hits when I click the button
        pullAndRenderCards(CARDS_PER_PULL);
      };
```

js/main.js
miffens: I'd prefer to keep what I have since it works, and to make it part of an HTML template I'd presume I have to do
```
      this.front = this.getElementsByClassName("card-front")[0];
      this.back = this.getElementsByClassName("card-back")[0];
```
except for all 10 card attributes, which sounds tedious orz

feel free to surface this in the discord channel!

js/main.js
```diff
@@ -55,12 +57,65 @@ async function defineCardComponent() {
     //Binds the card's onclick events to flip and show the description popup.
     setupOnClickEvents() {
       this.back.onclick = (event) => this.flipCard();
-      this.front.onclick = (event) => alert("TODO: Description popup.");
+      this.front.onclick = (event) => this.showDetails();
+    }
+
+    showDetails() {
+      updateDetailsDialog(this.data, this.getImageURL());
+      DETAILS_DIALOG_A11Y.show();
```
miffens, line 65: Sometimes I used the JS API (show), sometimes I added attribute tags to the HTML (hide) - if the HTML way didn't work I used JS and didn't dig too much into why the HTML way wasn't working :')

js/main.js
o5v3: Found the issue, it's on line 78 in main.js:

`document.body.innerHTML += html;`

When it's called it recreates the entire DOM anew, which invalidates our references declared at the beginning of the file:

`const GACHA_BUTTON = document.getElementById("gacha-button");`

So what happens is that the onclick event is set, and then the DOM is reconstructed, creating an entirely new button that doesn't have the onclick event set. This can be fixed by changing line 78 to:

`document.body.insertAdjacentHTML("beforeend", html);`

o5v3 (approved):

PR #12, opened 2022-12-19
Refactor Javascript into ES6 Modules

o5v3:


js/main.js
```diff
@@ -1,183 +1,50 @@
...
-  }
-}
-
 async function main() {
   await defineCardComponent();
-  getCSVAndMaybeRenderCollection();
+  getCSVData((cards_data) => {
+    if (CURRENT_PAGE == "/collection.html") {
+      renderCards(cards_data, COLLECTIONS_MAIN_CONTENT);
+    }
```
miffens, line 37: I like this callback change - consolidates the current page checking in one spot

pages/collection.html
```diff
@@ -58,11 +58,22 @@ <h1>The Regis Altare Birthday Project 2023</h1>
           Github</a
         >
       </footer>
+      <div
```
miffens: you probably didn't apply my recent changes, but just to be sure I moved the dialog code to a separate file

pages/collection.html
```diff
@@ -58,11 +58,22 @@ <h1>The Regis Altare Birthday Project 2023</h1>
...
+        <img class="details-dialog-card" />
+        <section class="details-dialog-text"></section>
+      </div>
     </main>
 
     <!-- Javascript must be at end so it can access the loaded DOM -->
     <script type="text/javascript" src="../js/papaparse.min.js"></script>
     <script type="text/javascript" src="../js/a11y-dialog.min.js"></script>
-    <script type="text/javascript" src="../js/main.js"></script>
+    <script type="module" src="../js/main.js"></script>
```
miffens, line 66: No objections from me for changing it to module - I don't know the difference between text/javascript and module in depth but from a super brief google it looks like module is best practice

PR #13, opened 2022-12-26
Set up pulls by rarity and card slots

o5v3:
Pokemon TCG packs use the concept of "slots"; these slots determine what kind of rarity can be found at a determined position when opening the pack. For example, the 1st slot may contain only common cards, the 2nd common and uncommon cards, and so on. This PR implements facilities to declare slots along with the probabilities for each rarity in each slot, along with a debug function (testRates) to check the average probabilities to get each rarity over time. 

With this we can also "nudge" rates a little to boost rarer cards should the need arise; this only works with rarities at the time, to do the same with individual cards (Reward a new card on X duplicates for example, like a pity system of sorts) it would be in a different PR.

### DIFF
--- a/js/gacha.js
+++ b/js/gacha.js
@@ -1,10 +1,94 @@
 import { renderCards } from "./cards.js";
+import { cards_by_rarity } from "./main.js";
 
 export const GACHA_BUTTON = document.getElementById("gacha-button");
-export const CARDS_PER_PULL = 10;
+//Card slots on each pull, representing the percent chance of getting a card of that rarity in each slot.
+//The sum of all rarities on a slot should be 100 or higher for proper function.
+const slots = [
+  {
+    Common: 100,
+    Uncommon: 0,
+    Rare: 0,
+    HoloRare: 0,
+    UltraRare: 0,
+    SecretRare: 0,
+  },
+  {
+    Common: 80,
+    Uncommon: 20,
+    Rare: 0,
+    HoloRare: 0,
+    UltraRare: 0,
+    SecretRare: 0,
+  },
+  {
+    Common: 0,
+    Uncommon: 100,
+    Rare: 0,
+    HoloRare: 0,
+    UltraRare: 0,
+    SecretRare: 0,
+  },
+  {
+    Common: 26,
+    Uncommon: 26,
+    Rare: 0,
+    HoloRare: 21,
+    UltraRare: 16,
+    SecretRare: 11,
+  },
+  {
+    Common: 0,
+    Uncommon: 0,
+    Rare: 40,
+    HoloRare: 20,
+    UltraRare: 20,
+    SecretRare: 20,
+  },
+];
+//Slots that replace one or more normal slots on special conditions.
+//This can be used to help collect all cards.
+const specialSlots = {
+  69: {
+    Common: 0,
+    Uncommon: 0,
+    Rare: 0,
+    HoloRare: 100 / 3,
+    UltraRare: 100 / 3,
+    SecretRare: 100 / 3,
+  },
+};
 
-//Pulls n cards from the passed "cards" array using the Fisher-Yates Shuffle.
-function pullCards(cards, n) {
+//Pulls a set of cards guided by an array of "slots" passed to the function.
+function pullCards(slots) {
+  let cards = [];
+  for (let slot of slots) {
+    let dice = Math.floor(Math.random() * 100) + 1;
+    if (dice in specialSlots) {
+      slot = specialSlots[dice];
+    }
+    for (let rarity in cards_by_rarity) {
+      if (dice <= slot[rarity]) {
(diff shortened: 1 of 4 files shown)


js/main.js
```diff
@@ -28,26 +29,31 @@ function getCSVData(callback = undefined) {
...
+      case "/gacha.html":
+        GACHA_BUTTON.onclick = (event) =>
+          pullAndRenderCards(cards_data, COLLECTIONS_MAIN_CONTENT);
+        await setupDetailsDialog();
+        break;
+      case "/collection.html":
+        await setupDetailsDialog();
+        renderCards(cards_data, COLLECTIONS_MAIN_CONTENT);
     }
+    getCardsByRarity();
```
miffens: (nit) `getCardsByRarity` feels more intuitive to me to be next to the cards_data setup in getCSVData. Move it there?

js/main.js
```diff
@@ -28,26 +29,31 @@ function getCSVData(callback = undefined) {
   });
 }
 
+function getCardsByRarity() {
```
miffens, line 42: (nit) add a comment explaining the shape of the returned result, ie
```
{ "Common": [ {card_name: "snailtare", rarity: "common",...}, {card}, {card}...]
  "Uncommon": [{card}],
   ....
}
```

js/gacha.js
```diff
@@ -23,8 +112,34 @@ function pullCards(cards, n) {
   return samples.slice(0, n);
 }
 
-//Pulls "n" number of cards from the cards_data array and renders them in CARD_LIST.
-export function pullAndRenderCards(cards_data, n, render_location) {
-  let pulled = pullCards(cards_data, n);
+//Pulls cards from the cards_data array and renders them in render_location.
+export function pullAndRenderCards(cards_data, render_location) {
+  //let pulled = getRandomCards(cards_data, n);
```
miffens: delete commented code

js/gacha.js
```diff
@@ -1,10 +1,99 @@
...
+};
+
+//Pulls a set of cards guided by an array of "slots" passed to the function.
+function pullCards(slots) {
+  let cards = [];
+  for (let slot of slots) {
+    let dice = Math.floor(Math.random() * 100) + 1;
+    if (dice in specialSlots) {
+      slot = specialSlots[dice];
+    }
```
miffens, line 75: does this actually occur in the real TCG rate system? :0

js/gacha.js
```diff
@@ -1,10 +1,99 @@
...
+    for (let rarity in slot) {
+      if (dice <= slot[rarity]) {
+        cards.push(getRandomCards(cards_by_rarity[rarity], 1)[0]);
+        break;
+      } else {
+        dice -= slot[rarity];
+      }
```
miffens: so the higher the dice value, the rarer the card?
Might be helpful to comment that to help readers understand the algorithm more easily

js/gacha.js
```diff
@@ -23,8 +112,34 @@ function pullCards(cards, n) {
...
+    }
+  }
+  for (let rarity in cards_by_rarity) {
+    rates[rarity] /= total;
+    rates[rarity] *= 100;
+  }
+  return rates;
+}
+
+function testRates() {
```
miffens: I didn't run this myself, but trust that you have o7

miffens (approved):

js/gacha.js
o5v3: Yes! Take a look at the official page for Booster Pack Slots: [https://tcg.pokemon.com/en-us/tcgl/droprate/](https://tcg.pokemon.com/en-us/tcgl/droprate/)
Slots
These are the slots for the "Sword & Shield-Lost Origin" expansion. The ones on the top row (Available Rarity) are the normal slots, and the ones on the bottom row (Available Upgrade Rarity) are special slots that have a small chance of replacing the standard drop in that slot. I imagine it brings more excitement to the pulls, and it also helps those who open them in getting rare cards.

o5v3:
Current average rates across 10.000 pulls:

Element: 9.902%
Common: 30.696%
Uncommon: 24.03%
Rare: 13.194%
HoloRare: 8.099%
UltraRare: 7.156%
SecretRare: 6.923%

js/main.js
o5v3: Moved there o7

js/main.js
o5v3: Done o7

js/gacha.js
o5v3: Done o7

js/gacha.js
o5v3: Yes, basically. It's a little more complex, but the comment should explain it better now, it's a pretty long comment lol

js/gacha.js
o5v3: Ran it again and left the results in a comment below o7

PR #14, opened 2022-12-28
Nav CSS & responsive layout for narrow views

bonbombs:
HTML Updates:

- Adding credits & artist/writer's message board HTML as placeholder pages
- Separated out the required nav items into a component that can be edited in `nav.html` so it's easier to make edits to across all pages
- Replaced existing nav items with `<app-nav>` tag element 

CSS Updates:

- Additional styling to nav tabs
- Nav elements has a media query check so tabs shrink into a menu when window width is <1000px

JS Updates:
- Added a `Nav` component that populates on every page and has the necessary click events

Other:

- Added an `icons/` folder to hold svg files. If they should be with images instead, I'm happy to move those over.


bonbombs:
@miffens @o5v3 I wonder if y'all have any thoughts on my approach w/ dynamically loading in the nav elements. Upon checking the deploy preview, there seems to be a noticeable load before seeing the nav elements populate in. Figuring out another way to do it in a preprocess step would make sense but it would eat into time that can be used elsewhere in the project. I'm happy to revert back to hard-coding the nav for performance/load time reasons if that approach makes the most sense.

o5v3:
Indeed, there's a split second between the page load and the nav appearing. Hard-coding it is viable since we're only using a few pages, but the structure is clearer and more organized in the dynamic way... I propose to hard-code it for now as it's only rendered one time per page. If there comes a situation where we need to place more than one per page, change its location, etc, then we could change to dynamic then. What do you think, @miffens ?

Oh btw, when I open the dialog popup, the active tab appears on top of the popup. I think that's a bug.

miffens:
I agree with leaving the nav hardcoded and be extra thorough with manual QA with making sure the nav is consistent across pages. I also noticed the background is white for a moment before it turns dark blue. Could that white flash also be addressed, presuming it's a small CSS change?

miffens:
also, what is the DS_store file?

bonbombs:
oops... looks like DS_store was generated by my mac. I can add that file to the gitignore and remove it from the repo. 

Good catch on the bugs, I'll address those. thank you!

collection.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
```
miffens: Is this an extra file that should be deleted? the `pages/collections.html` is the one we're using I believe?

css/welcome.css
```diff
@@ -112,7 +157,7 @@
     left: 0;
     width: 100vw;
     height: 100vh;
-    background-color: var(--accent-color);
+    /* background-color: var(--accent-color); */
```
miffens: delete commented code?

css/card.css
```diff
@@ -59,6 +59,7 @@
   background-color: #fff;
   height: 90%;
   width: 90%;
+  z-index: 10;
```
miffens: (nit) maybe comment explaining the z-index numbers, ie "above nav a.active" (or whatever would appear above it if this rule weren't specified)

css/main.css
```diff
@@ -40,10 +44,15 @@ footer {
...
 
+footer a {
+  color: white;
+}
+
 footer svg {
   width: 1em;
 }
 
+
```
miffens: (super nit) extra space

nav.html
```diff
@@ -0,0 +1,42 @@
+<!-- CSS include -->
```
miffens: delete now that we're back to hardcoding?

pages/message-board.html
```diff
@@ -70,5 +106,8 @@ <h1>The Regis Altare Birthday Project 2023</h1>
         >
       </footer>
     </main>
+    <script type="text/javascript" src="../js/papaparse.min.js"></script>
+    <script type="text/javascript" src="../js/a11y-dialog.min.js"></script>
+    <script type="module" src="../js/main.js"></script>
```
miffens: we probably won't need papaparse for the message board. Unless it needs to be included because main.js requires it?

pages/message-board.html
```diff
@@ -35,13 +34,50 @@
     <main>
       <h1>The Regis Altare Birthday Project 2023</h1>
 
+      <a href="#message-board" class="visually-hidden">Skip to content</a>
```
miffens, line 37: thanks for adding the skip link!

pages/message-board.html
```diff
@@ -35,13 +34,50 @@
...
+      <svg width="0" height="0" aria-hidden="true">
+          <clipPath id="navClip" clipPathUnits="objectBoundingBox">
+              <path transform="scale(0.004388274530454626, 0.019230769230769232)" d="M173.88,0H54C38,0,45,21,0,52H227.88C182.88,21,189.88,0,173.88,0Z"/>
+          </clipPath>
+      </svg>
+
+      <!-- Burger menu for responsive layout -->
+      <button id="nav-button" aria-label="Toggle site navigation">
+          <span class="visually-hidden">Menu</span>
+          <img class="icon icon-menu" aria-hidden="true" src="../icons/Icons_Hamburger.svg"/>
```
miffens: (nit) If the button has aria-label, I think that will take precedence over the button contents, so I don't think the visually hidden text and aria-hidden is needed
I'm not on my Mac rn so can't confirm

pages/message-board.html
```diff
@@ -35,13 +34,50 @@
...
-        <a href="../pages/collection.html">Collection</a>
-        <a href="../pages/gacha.html">Gacha</a>
-        <a class="active" href="../pages/message-board.html">Birthday Message Board</a>
+          <a id="nav-home" href="../index.html">
+            <img class="icon icon-logo" aria-hidden="true" src="../icons/Icons_Open-Packs.svg"/>
+            <span>Home</span>
+          </a>
+          <a id="nav-gacha" href="../pages/gacha.html">
+              <img class="icon icon-gacha" aria-hidden="true" src="../icons/Icons_Open-Packs.svg"/>
+              Gacha</a>
```
miffens, line 59: I didn't read the new nav css too closely so can't pinpoint the cause, but I noticed while tabbing around that the keyboard focus highlight is dark blue/ hard to see on the nav items specifically, while other buttons have the default black/ white stroked highlight. Would this be a quick fix?

js/main.js
```diff
@@ -31,6 +31,31 @@ function getCSVData(callback = undefined) {
...
+ */
+function setupNavClickEvents () {
+  let mobileButton = document.getElementById("nav-button");
+  let closeButton = document.getElementById("close-nav-menu");
+  let navElement = document.getElementsByTagName("nav")[0];
+  function openNavMenu (event) {
+    this.classList.add("active");
+    setTimeout(() => {
+      this.classList.add("open");
+    }, 250);
```
miffens: add a comment explaining the reason a timeout is needed, and if the 250 number is significant?

miffens (commented):
Not a full review, will dig into the CSS in more depth later

collection.html
bonbombs: yep, good catch. I can remove that

pages/message-board.html
bonbombs: yeah that was the reason why I included papaparse. I could separate the nav js event handlers into a separate file to reduce the amount of stuff included for pages that don't need papaparse.

pages/message-board.html
bonbombs: Did a quick test and can confirm that the aria-label takes precedence. Will remove the redundant span tag!

pages/message-board.html
bonbombs: Oh jeez, good catch! Looks like applying `clip-path` on the `a` tags was causing the keyboard focus outlines to not render. I'll move the `clip-path` styling to a ::before psuedo element.

css/nav.css
```diff
@@ -1,21 +1,233 @@
...
+section#message-board {
+    background-color: #485EB3;
+}
+
+nav a#nav-credits::before,
+section#credits {
+    background-color: #2852B3;
+}
+
+#nav-button {
```
miffens: (nit) I believe best practice is to have id's be unique. Should this be a class?

Don't worry if this is too tedious to refactor

css/nav.css
```diff
@@ -1,21 +1,233 @@
...
+        clip-path: url(#navClip);
+
+        z-index: -1; /* render underneath text */
+    }
+
+    nav a:first-of-type {
+        transform-origin: 0 100%;
+    }
+
+    /* specifying #nav-credits here since the last element in the nav is not a tab */
```
miffens, line 225: could you explain this more? what does "not a tab" mean? I didn't read the code super closely and might have missed this

pages/message-board.html
miffens: If it wouldn't take much time, that sounds better!

miffens (approved):
Approving what's currently here. You can merge what you have and add the new assets in a different PR

css/nav.css
bonbombs: yeah! Originally I had `:last-of-type` as the selector however the selector is not targeting the appropriate tab (`#nav-credits` in this case) because it's not the last child element in `nav`. The last child element is the close button that we need in the hamburger menu.

css/nav.css
bonbombs: Probably could be named better for sure. It's the hamburger menu button so there's probably only one instance of it. I'll rename it to be something like `#nav-hamburger`

PR #16, opened 2022-12-28
Save pulled cards and choose to show full collection or only owned in collection page

o5v3:
- Saves pulled cards in localStorage, which can be used as a key-value store for checking owned cards.
- Sets up a toggle in collection page to choose whether to show only owned cards or the full collection (with cards not owned darkened and in grayscale).
- There's also a button to reset localStorage to its default state. Could change it out later if we end up storing more information than just cards and sorting type in there.
- Also implements an easy way to declare sorting types and functions. Declare your sorting function inside "sort_functions", and then call sortCards([cards_to_sort], sort_type) to get your sorted array. 

Note: This saves owned cards in localStorage, so it persists even if you turn off your device and turn it on again. If we want to make it ephemeral, it can be changed to sessionStorage which uses the same API but erases data when you close all browser tabs. Data doesn't persist between devices as of now, to do that we would need to set up some sort of basic authentication, a database and possibly use Netlify Functions as a serverless backend. Don't think we need to do that though, as what we have with localStorage should be more than enough for our needs.

### DIFF
--- a/js/cards.js
+++ b/js/cards.js
@@ -115,10 +115,11 @@ export function showCollection(cards_data, htmlLocation) {
   } else {
     let ownedCards = getOwnedCards(cards_data);
     if (ownedCards.length == 0) {
-      alert("TODO: No cards to show");
+      htmlLocation.innerHTML =
+        "You have no cards at the moment. Try pulling some at the gacha!";
     } else {
       ownedCards = sortCards(ownedCards, sort);
+      renderCards(ownedCards, htmlLocation, true);
     }
-    renderCards(ownedCards, htmlLocation, true);
   }
 }
--- a/js/cards.js
+++ b/js/cards.js
@@ -3,6 +3,18 @@ import { updateDetailsDialog, DETAILS_DIALOG_A11Y } from "./dialog.js";
 
 const CLOUD_NAME = "dazcxdgiy";
 const CLOUDINARY_URL = `https://res.cloudinary.com/${CLOUD_NAME}/image/upload/`;
+const COLLECTED_CARDS_NUMBER = document.getElementById(
+  "collected-cards-number"
+);
+const RARITIES = [
+  "Element",
+  "Common",
+  "Uncommon",
+  "Rare",
+  "HoloRare",
+  "UltraRare",
+  "SecretRare",
+];
 
 //Object to store all supported sorting functions.
 //Can be used by calling:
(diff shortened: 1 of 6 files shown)


js/cards.js
```diff
@@ -70,3 +84,42 @@ export function renderCards(cards, htmlLocation, replace = false) {
...
+
+function getOwnedCards(cards_data) {
+  let ownedCards = [];
+  let storage = { ...localStorage };
+  for (let item in storage) {
+    if (item.slice(0, 4) === "card") {
+      let card_id = item.split("-")[1];
+      ownedCards.push(
+        cards_data.find((card) => card["Collector Number"] === card_id)
+      );
```
miffens, line 104: I wonder how the performance of this will hold up when we have hundreds of cards. Maybe it'd be worth having a `cards_by_collector_number` object alongside the rarity one. That is a memory tradeoff though.

Another idea is to push the whole card object to localStorage. This would also simplify more detailed statistics like # of cards collected by rarity (ie 10 common, 2 rares...)

miffens (approved):
I'm good with this as the first iteration of the collections feature. I agree localStorage suffices and we don't need to get fancier.

One suggestion I think would be quick to add is a "x/total collected" stat text visible in both the gacha and collection views.

Some bigger suggestions that can be done in a followup are:
 - Change the darkened cards to either be fully obscured or fully visible. Perhaps instead of a "show full collection" button it would become a dropdown with the options "just your cards", "show your cards alongside uncollected cards", "show all cards (spoilers!)"
 - Also track total pulls + ability to render past pulls

js/cards.js
o5v3: That's a good idea, let me write a reminder to do that if performance starts to suffer. About pushing it all to localStorage, I feel like it's a little overkill given that we already have all the info in cards_data, cards_by_rarity and potentially cards_by_collector_number, each of them using references to the same card objects, while pushing it to localStorage would bring another array but with brand new references not shared with the others (localStorage saves strings only, so the objects would have to be rebuilt on retrieval). What we can do is store preprocessed data in localStorage to save time. I already made a commit to save card count by rarity in localStorage and keep track of pulls, we can add more things to it if, again, performance starts to suffer.

o5v3: Also took away the "darkened card" overlay. It's gone now, poof.

PR #18, opened 2022-12-31
Artist/Writer Messages Page Setup

bonbombs:
Styling still pending.

Still need to setup CSV pull. Right now the messages are being generated by a `for` loop.

Also need to figure out how to deal w/ VoiceOver not reading the article messages.


bonbombs:
@miffens I edited the PR title since I'm not sure what the CSV is going to look like yet. Maybe we merge this in w/ the initial setup, the fixes to the tabs, and the dark reader fix?

I also wasn't able to get VoiceOver to read the contents of the article elements. As a sanity check, I tried Chrome's screen reader extension and that seemed to work. Am I missing an aria tag here that VoiceOver looks for or should I use a different semantic tag that more compatible to define each message element?

css/main.css
```diff
@@ -110,8 +113,40 @@ select:hover {
...
+article.header-message {
+  column-span: all;
+  text-align: center;
+  margin-bottom: 1em;
+}
+article.artist-writer-message {
+  padding: 1.5em;
+  background: white;
+  border-top-left-radius: 0.5em;
+  border-bottom-right-radius: 0.5em;
```
miffens, line 143: I assume the border radius not on every corner is intentional

css/main.css
```diff
@@ -110,8 +113,40 @@ select:hover {
...
+  margin-top: 0;
+  margin-bottom: 0.5em;
+}
+
 @media only screen and (max-width: 600px) {
   /* mobile / portrait styling */
+
+  section#artist-writer-board {
+    column-count: 1;
+  }
```
miffens: (nit) imo the transition to 1 col could happen like 100-200px sooner. The 2 cols get very narrow leading up to this breakpoint

css/nav.css
```diff
@@ -200,7 +230,7 @@ section#credits {
 
     nav a:not(#nav-home) {
         padding: 0.55rem 2.7rem;
-        transform-origin: 50% 100%;
+        transform-origin: 50% 0%;
```
miffens: (highlighted CSS unrelated to comment) while you're here, would you mind polishing our buttons a bit:
 - the 'open pack' button doesn't change colors on hover
 - the hover color doesn't appear if you are keyboard focused on the button

miffens (approved):

miffens:
Voiceover reads "article" and "end of article" for me. I also think it's fine in case screenreaders don't - I think a user can understand the page contents fine without those annotations
(nit) The message headings should ideally be h2-h3 - currently the heading hierarchy jumps from h1 ("the birthday project" title) to h4-h5 without reason. imo it's not a huge problem - it wouldn't confuse a user much because the messages being subcontents of the page is effectively conveyed with h4-h5, it's just slightly unexpected lol.
(nit) this is slightly fancy so not needed and mainly for your learning - if the messages are wrapped in ul/li elements, Voiceover will read the "position" you are on in the list ("2 of 50"). This can be helpful to let the user know how much more content there is remaining.

css/main.css
bonbombs: yep! that's correct

bonbombs:
> Voiceover reads "article" and "end of article" for me. I also think it's fine in case screenreaders don't - I think a user can understand the page contents fine without those annotations (nit) The message headings should ideally be h2-h3 - currently the heading hierarchy jumps from h1 ("the birthday project" title) to h4-h5 without reason. imo it's not a huge problem - it wouldn't confuse a user much because the messages being subcontents of the page is effectively conveyed with h4-h5, it's just slightly unexpected lol. (nit) this is slightly fancy so not needed and mainly for your learning - if the messages are wrapped in ul/li elements, Voiceover will read the "position" you are on in the list ("2 of 50"). This can be helpful to let the user know how much more content there is remaining.

@miffens I made attempts to wrap the messages in ul/li but Voiceover doesn't seem to want to say which position list item I'm on in the list. Not sure if I'm doing something wrong in the code or if I misconfigured my Voiceover. Added hover states for when a user highlights a button w/ a keyboard!

https://user-images.githubusercontent.com/7121716/211129266-ed27aa3a-bd14-4541-a324-1465b98534ee.mov

js/messages.js
```diff
@@ -0,0 +1,51 @@
...
+    let messageContainer = document.getElementById("artist-writer-messages");
+    for (var i = 1; i < 100; i++) {
+      let messageNode = document.createElement("li");
+      let messageArticle = document.createElement("article");
+      // TODO: Dummy data for now. Alternate roles for now as a test. This should be pulled from a csv somewhere.
+      let userRole = i % 2 === 0 ? "Artist" : "Writer";
+      let userSocial = `user_social_${i}`;
+      let h3String = [userRole, userSocial].join(" | ");
+      messageNode.tabIndex = 0;
+      messageArticle.tabIndex = 0;
```
miffens: Remove - keyboard focus is only for interactive elements like buttons and links. A keyboard user can use arrow keys to scroll the page. Excessive tabbable elements will keep keyboard users from reaching the interactive ones

Would this happen to be interfering with the list position being read? If that's not it, my guess is the nesting of article in an li is unconventional. I would feel okay with removing the article tag nesting for just li if I felt list position was important to convey to a user. If I had more resources I would test on real users and go with the option they say prefer, but in the absence of that, using one or the other of li or article is reasonable to me.

bonbombs: I **just** realized I was using voiceover wrong LOL. Was tapping the tab key and not using the VO-modifier w/ arrow keys. Sorry about that. Removed the tabIndexes. I am now hearing/seeing voiceover narrate each li with its appropriate position.

PR #19, opened 2022-12-31
Card Animations for Gacha and Dialog Pages

kirinarimasu:
Testing out applying the card animation to the current site layout/format.


js/dialog.js
```diff
@@ -28,9 +29,29 @@ export function updateDetailsDialog(data, cardUrl) {
...
+  // Card art is applied
+  DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0].style.backgroundImage = 'url("' +
+    cardUrl + '")';
+
+  // Add animation start
+  DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0].classList.add(cardRarity(data["Rarity Folder"]), "animated");
+  
+  // Animation reset
+  DETAILS_DIALOG_EL.getElementsByClassName("card-details-dialog-close")[0].addEventListener("click", cardReset);
+  document.getElementsByClassName("dialog-overlay")[0].addEventListener("click", cardReset);
```
miffens: I think you want `dialog.on('hide',` here. The dialog can also close on esc and maybe other controls too. Best let the a11ydialog library handle all that and hook into the hide event listener it provides.
https://a11y-dialog.netlify.app/advanced/events

Is the consequence of not having this code at all that if the user closes, then reopens the dialog with a different card, the different card could have classes from the previous? Like, "animated animated rare ultra"

js/dialog.js
```diff
@@ -28,9 +29,29 @@ export function updateDetailsDialog(data, cardUrl) {
...
+  let dialogCard = DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0];
+  dialogCard.addEventListener("animationstart", function() {
+    dialogCard.removeEventListener("mouseover", cardAnimate);
+    dialogCard.removeEventListener("click", cardAnimate);
+  }, false);
+  dialogCard.addEventListener("animationend", function() {
+    dialogCard.classList.remove("animated");
+    dialogCard.addEventListener("mouseover", cardAnimate);
+    dialogCard.addEventListener("click", cardAnimate);
+  }, false);
```
miffens: fine to keep as-is since it works, but I was wondering if only adding/ removing the "animated" class suffices? Would this still work without adding/ removing the event listeners on each run of the animation? I'm not sure if adding/ removing event listeners so frequently is "best practice" - @bonbombs or @o5v3 might know more

js/dialog.js
```diff
@@ -48,3 +69,30 @@ export function updateDetailsDialog(data, cardUrl) {
   `
   );
 }
+
+function cardAnimate(dialogCard){
```
miffens: (nit) naming functions that "do" something like `[verb][subject]` helps me with reading code in a plain English style, so:
`animateCard`
`resetCardEffects`
`setCardRarity`

"on mouseover, card animate" > "on mouseover, animate card"

pages/gacha.html
```diff
@@ -77,8 +77,18 @@ <h1>The Regis Altare Birthday Project 2023</h1>
...
+          <input type="radio" id="gacha-pile" name="gacha-display" value="gacha-pile">
+          <label for="gacha-pile"> Card Pile</label>
+          <input type="radio" id="gacha-grid" name="gacha-display" value="gacha-grid">
+          <label for="gacha-grid"> Card Grid</label>
+        </form>
+        <button id="gacha-button" type="submit">Pull Gacha</button>
+        <div id="gacha-notes">
+          <p>Re-pulling while half viewed, or flipping a card in 'pile' view will not allow you to see previous cards, but will be saved to your Collection. When the card has stopped moving, animate it again by clicking / hovering!</p>
+        </div>
+        <section id="card-list">Ain't nobody here but us Altventurers. Pull the gacha to get started!</section>
```
miffens, line 91: How about moving the "Pull Gacha" button closer to where the cards render/ this "Pull the gacha to get started" note?
Also, how about renaming "Pull Gacha" to "Open Card Pack"? so the language is more in the TCG style?

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+}
+#gacha-controls tcg-card {
+  --card-width: 400px;
+  --card-height: 559px;
+  position: absolute;
+  /* margin-left: auto;
+  margin-right: auto;
+  left: 0;
+  right: 0; */
+  /* right: calc(100vw - (90px + 1em + var(--card-width))); */
```
miffens: delete commented out code throughout this file

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+  display: inline-block;
+}
+tcg-card {
+  margin: 3%;
+  perspective: 600px;
+}
+
+#gacha-controls #card-list {
+  min-height:52vh;
+}
```
miffens: I think this is what's causing the cards to sometimes clip out of the container at certain dimensions:
Screenshot 2023-01-03 123826
We can address this in a different PR though - I'm excited to get these animations merged!

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+  position: absolute;
+  /* margin-left: auto;
+  margin-right: auto;
+  left: 0;
+  right: 0; */
+  /* right: calc(100vw - (90px + 1em + var(--card-width))); */
+  right:10%;
+  transition: all 0.3s ease;
+}
+tcg-card .card-info {
```
miffens: (nit) how about the name card-subtitle instead? Helps me picture what this class corresponds to more clearly

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+  font-size:0.75em;
+  color:white;
+  margin-top:2%;
+}
+#gacha-controls tcg-card .card-info {
+  display:none;
+}
```
miffens: (nit) ideally we would not add the "card info" to the DOM at all for the gacha view, but we can clean this up in a future PR if we have time

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+
+#gacha-controls tcg-card:nth-child(3n+1){
```
miffens: (nit) comment that this CSS is for the "pile"

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+  transform-style: preserve-3d;
+  transition: box-shadow .4s ease;
+  transform-origin: center;
+  /* box-shadow: 0 10px 20px -5px #000; */
+  box-shadow: 0px 10px 20px -5px rgba(0, 0, 0, 0.3);
+  border-radius: var(--border-radius);
+}
+.card-component:hover {
+  box-shadow: 0px 10px 20px -5px rgba(0, 0, 0, 0.3);
+}
```
miffens: redundant with box-shadow on the card-component itself right now. Should we adjust this so that the shadow does change on hover? Maybe have it light up white?
Can be a small polish PR later so it doesn't slow this one down - if so, delete this hover rule for now

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+#gacha-controls .card-component {
+  box-shadow:none;
+}
+
+#gacha-controls .card-front {
+  backface-visibility: visible;
+  will-change: transform;
+  transform: rotateY(180deg) translateZ(1px);
```
miffens: I don't understand these rules well. Would be great for @bonbombs to review as well, but no need to block this merge on that

css/card.css
```diff
@@ -1,30 +1,163 @@
...
+  animation: flip 0.5s ease 0s 1;
+  animation-fill-mode: forwards;
+}
+@keyframes flip {
+  0% {
+    transform: rotate3d(0, 1, 0, 0deg);
+  }
+  100% {
+    transform: rotate3d(0, 1, 0, -180deg);
+  }
```
miffens: what's the difference between this and
`#gacha-controls #card-list tcg-card.clicked` above? I can't really tell just from the class/ id names alone. Maybe add a comment explaining the difference?

css/card.css
```diff
@@ -43,8 +176,92 @@
...
+}
+#card-details-dialog::-webkit-scrollbar-track {
+  border-radius:0px 10px 10px 0px;
+}
+.details-dialog-text::-webkit-scrollbar-thumb, #card-details-dialog::-webkit-scrollbar-thumb {
+  background-color: var(--bg-color);
+  border: 3px solid var(--accent-color);
+  border-radius:10px;
+}
+
```
miffens: I think we can delete this and not go out of our way to style the dialog scrollbar (unless Bing really wanted this and I missed the message) - we're likely to miss some browsers

css/card.css
```diff
@@ -66,28 +283,104 @@
...
+  overflow-y: auto;
 }
 
 [role="dialog"][aria-hidden="true"] {
   display: none;
 }
 
+[role="dialog"][aria-hidden="noshow"] {
+  opacity: 0;
+}
```
miffens: what's the difference between this opacity: 0 and the existing display: none rule?

css/card.css
```diff
@@ -66,28 +283,104 @@
...
+
   #card-details-dialog {
     grid-template-areas:
       ".     x"
       "head  head"
       "text  text"
       "card  card";
-    overflow: auto;
+    overflow-y: auto;
+    overflow-x: hidden;
```
miffens: I worry overflow-x might hide something that shouldn't be hidden. Maybe back out these changes and let @bonbombs focus on it

css/card.css
```diff
@@ -66,28 +283,104 @@
...
+    }
+    100% {
+      margin-top: calc(var(--card-height) + 3% + 100px);
+    }
+  }
+}
+
+@media screen and (max-width: 550px) {
+  #gacha-display-toggle {
+    float:none;
```
miffens: Is float: none needed? I can't find anywhere that is setting float

css/card.css
```diff
@@ -101,3 +394,345 @@
...
+  bottom: 0;
+  top: 0;
+  background-repeat: no-repeat;
+  opacity: 0;
+  mix-blend-mode: color-dodge;
+  transition: all 0.33s ease;
+  border-radius: 5%/3.5%;
+}
+
+.details-dialog-card:before, #gacha-controls .card-image:before {
```
miffens, line 466: Add comments throughout below, particularly calling out important lines for common things to update, like higher brightness/ contrast, different texture?

css/card.css
```diff
@@ -101,3 +394,345 @@
...
+  background-position: center;
+  background-size: 200%;
+  background-blend-mode: overlay;
+  mix-blend-mode: screen;
+}
+
+.details-dialog-card.holo:after, #gacha-controls .card-image.holo:after {
+  --space: 4%;
+  --galaxybg: center 521px;
+  background-image: url(../images/galaxy.jpg), url(../images/galaxy.jpg), url(../images/galaxy.jpg), repeating-linear-gradient( 82deg, rgb(219, 204, 86) calc(var(--space)*1), rgb(121, 199, 58) calc(var(--space)*2), rgb(58, 192, 183) calc(var(--space)*3), rgb(71, 98, 207) calc(var(--space)*4), rgb(170, 69, 209) calc(var(--space)*5), rgb(255, 90, 180) calc(var(--space)*6), rgb(255, 90, 180) calc(var(--space)*7), rgb(170, 69, 209) calc(var(--space)*8), rgb(71, 98, 207) calc(var(--space)*9), rgb(58, 192, 183) calc(var(--space)*10), rgb(121, 199, 58) calc(var(--space)*11), rgb(219, 204, 86) calc(var(--space)*12) );
```
miffens, line 506: I'm especially curious how to read this line. How were the numbers chosen? How are they affected when the card moves in 3d?

miffens (commented):
All small things + "how does this work" questions

css/card.css
miffens: ah, o5v3 covered this in #22

css/card.css
bonbombs: `will-change` is an interesting property and there's a bunch of nitpicky ways to better use it to increase optimization. Ideally,`will-change` should be applied right before the animation fires in JS so the browser knows to optimize the performance for that particular animation. For now though I think it's fine. Here's an article that lays it out in detail: https://dev.opera.com/articles/css-will-change-property/

The rest looks good though!

css/card.css
bonbombs: I agree. I think leaving it at `overflow: auto` is sufficient. afaik, there shouldn't be any elements that are causing the dialog container to expanding past its defined width.

js/dialog.js
kirinarimasu: I will need to revisit this later, I just had a go at swapping this out and it no longer works as expected. Yes, the animation relies on the appropriate classes being added to a the particular div that holds the card image; and this will hold on to the previous styles when the dialog is closed if not reset (feedback from Bing previously).

js/dialog.js
kirinarimasu: I'm not sure if it's really best practice or not, but it affects the experience; as the animation will keep stopping and starting. What I'm trying to achieve is a full run of the animation without interruptions from the user; then the opportunity to start it again. :)

bonbombs: It looks like the goal here is that you want to prevent the animation from restarting if a user tries to click on the card mid-animation, right? 
You can probably add a class to the dialogCard called `unclickable` and in CSS, define `.unclickable { pointer-events: none; }`. That way we can prevent the click events from firing on animation, while still retaining the event listeners.

js/dialog.js
kirinarimasu: Done, I think. Renamed those functions.

css/main.css
```diff
@@ -57,6 +57,21 @@ footer svg {
   text-shadow: 4px 4px 1px rgb(13 66 163), -3px -3px 1px rgb(148 186 255);
 }
 
+#welcome-splash-button {
```
bonbombs, line 60: ty for styling the buttons!! much appreciated!!

pages/gacha.html
kirinarimasu: Wording update is totally fine - this was just what was there by default and hadn't been changed.

The 'pull' button and the 'pull this gacha to get started' note are separated because the gacha then fully replaces it. Ideas I was playing with was hiding the light blue box altogether (so the text must be placed elsehwere); and only showing it when gacha is pulled.

css/card.css
kirinarimasu: When you display:none something you can't affect it until it displays as something again, even if it's still hidden. I actually removed whatever this was applied to and forgot to delete this sorry.

css/card.css
kirinarimasu: That's fine - there was actually a horizontal scroll coming up when resizing the window down; and I'm not sure why; so I just added that. But since you are updating further the layout etc I can remove.

css/card.css
kirinarimasu: Sorry, again something I tried and then deleted and forgot to remove from here too.

css/card.css
kirinarimasu: Removed the commented out code.

css/card.css
kirinarimasu: These numbers are taken directly from the simey stuff; so the specifics of why they are chosen I don't really know. It's just creating a layered background (images and then gradients), which then uses a number of blend effects. It's easier to see the gradient by itself when there is no image applied.

js/dialog.js
kirinarimasu: Ahh just realised because I changed the type of listener I don't think it needs to be done that way anymore; so have updated it.

css/card.css
kirinarimasu: I added some comments throughout the CSS here, but I'm not sure they're as detailed as you would like.

css/main.css
kirinarimasu: It is very basic so feel free to change it. :)

miffens:
What I did: ran `git pull --rebase` and resolved the following merge conflicts:

Changed card-image to a div and kept the new comment
merge conflict 1

Not the best screenshot, but this was a downstream conflict from the rarity sorting that got added to this file in another PR alongside the card pile/ flip stuff in this PR. Accepted the blue highlighted stuff as well as keeping the rarity sorting block of code (not in this screenshot)
merge conflict 2

Once the merge conflict was resolved, I ran `git stage card.html` (or whatever file had the conflict), then `git rebase --continue` until it gave me another conflict to resolve, or the rebase completed successfully.
In this screenshot, you see:
 - I got a merge conflict, ran `git status` to see which file was the issue (in red)
 - staged the file once I fixed the conflict
 - continued the rebase
 - the rebase advanced to step 3/13 and had another conflict

I repeat the above until I complete 13/13
rebase example

Then when I tried to `git push` I got a `Updates were rejected because the tip of your current branch is behind` error, which from experience I know I can ignore - I messed with the commit order and github doesn't know how to reconcile that. So I do `git push -uf origin kiri/card-animations` (extra "f" next to the -u to mean "force") and all the commits are added to the end of the PR

card.html
```diff
@@ -1,9 +1,10 @@
-<div class="card">
+<div class="card-component">
   <button class="card-front">
-    <img class="card-image" />
+    <div class="card-image"></div>
```
miffens, line 3: I assume img was changed to div for the CSS effects to work?

css/card.css
```diff
@@ -1,30 +1,143 @@
...
+  transition: all 0.3s ease; /* make any changes to this card flow, so it doesn't snap */
+}
+tcg-card .card-info {
+  font-size:0.75em;
+  color:white;
+  margin-top:2%;
+}
+#gacha-controls tcg-card .card-info {
+  display:none; /* don't want to see that particular card information on the gacha page */
+}
```
miffens: can you delete these changes/ add your nice styling to o5v3's version of the card subtitle stuff?

css/card.css
miffens: they're good, thank you!

js/cards.js
```diff
@@ -66,13 +71,19 @@ export async function defineCardComponent() {
...
       );
       this.innerHTML = html;
+      this.holder = this.getElementsByClassName("card-component")[0];
       this.front = this.getElementsByClassName("card-front")[0];
       this.back = this.getElementsByClassName("card-back")[0];
-      const image = this.getElementsByClassName("card-image")[0];
-      image.src = this.getImageURL();
+      this.info = this.getElementsByClassName("card-info")[0];
+      this.image = this.getElementsByClassName("card-image")[0];
+      // const image = this.getElementsByClassName("card-image")[0];
```
miffens: delete commented code

pages/gacha.html
miffens: ok, sounds like it's not a super quick change to move the pull button, so I've added it to our Notion board as a nice to have

js/dialog.js
```diff
@@ -28,9 +30,28 @@ export function updateDetailsDialog(data, cardUrl) {
...
+
+  // animation events
+  let dialogCard = DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0];
+  dialogCard.addEventListener("animationend", function() {
+    dialogCard.classList.remove("animated");
+    if(first_open){
+      dialogCard.addEventListener("mouseover", animateCard);
+      dialogCard.addEventListener("click", animateCard);
+      first_open = false;
+    }
```
miffens: could this be in `setupDetailsDialog` so that the global variable isn't needed?

js/cards.js
```diff
@@ -85,15 +96,41 @@ export async function defineCardComponent() {
...
-      this.back.classList.toggle("hidden");
+
+      if(gacha_display_selection !== "gacha-grid" && previous_card_front !== undefined){
+        previous_card_front.removeEventListener("mouseover", addAnimation);
+        previous_card_front.removeEventListener("click", addAnimation);
+      }
+
+      this.classList.add("clicked");
+      this.holder.classList.add("flip");
+      previous_card = this;
```
miffens: what is the purpose of `previous_card`? I don't see it used elsewhere in this file. Delete?

js/cards.js
```diff
@@ -85,15 +96,41 @@ export async function defineCardComponent() {
...
+      this.back.classList.remove("hidden");
+      this.front.tabIndex = "-1";
+
+      addHover();
+    }
+
     flipCard() {
-      this.front.classList.toggle("hidden");
-      this.back.classList.toggle("hidden");
+
```
miffens: (nit) took me a minute to understand this code - a comment like "remove event listeners from previous flipped card and flip over + set up event listeners for current card" would help

js/cards.js
```diff
@@ -48,6 +48,11 @@ const sort_functions = {
       }
     },
 };
+let card_count = 10;
```
miffens: rename to something like `card_z_index` so it's clearer what this is for

js/cards.js
```diff
@@ -85,15 +96,41 @@ export async function defineCardComponent() {
...
 
     //Binds the card's onclick events to flip and show the description popup.
     setupOnClickEvents() {
       this.back.onclick = (event) => this.flipCard();
-      this.front.onclick = (event) => this.showDetails();
+      if(CARD_ART_HIDDEN_ON_LOAD){
+
+      } else {
+        this.front.onclick = (event) => this.showDetails();
+      }
```
miffens: condense to `if(!CARD_ART_HIDDEN_ON_LOAD)`?

js/cards.js
```diff
@@ -114,6 +151,47 @@ export async function defineCardComponent() {
...
+  let cardList = document.getElementsByClassName("clicked");
+  let el;
+  for (let i = 0; i < cardList.length; i++) {
+    let z = window.getComputedStyle(cardList[i]).zIndex;
+    el = cardList[i].children[0].children[0].children[0];
+    if(parseInt(z) !== card_count){
+      el.removeEventListener("mouseover", addAnimation);
+      el.removeEventListener("click", addAnimation);
+    } else {
+      console.log(el);
```
miffens: delete console.log

js/cards.js
```diff
@@ -114,6 +151,47 @@ export async function defineCardComponent() {
...
+export function addHover(){
+  let cardList = document.getElementsByClassName("card-image");
+  for (let i = 0; i < cardList.length; i++) {
+    cardList[i].addEventListener("animationend", function(){
+      cardList[i].classList.remove("animated");
+    })
+    cardList[i].addEventListener("mouseover", addAnimation);
+    cardList[i].addEventListener("click", addAnimation);
+  }
+}
```
miffens: the card flip animation currently lags a lot on my laptop - Is it just me? I wonder if adding so many event listeners has anything to do with it?

js/cards.js
```diff
@@ -85,15 +96,41 @@ export async function defineCardComponent() {
       return `${CLOUDINARY_URL}/q_auto/${this.data["Rarity Folder"]}/${this.data["Filename"]}.png`;
     }
 
+    resetCard() {
```
miffens: I think `setupCardPile` would be a better name for this function?
I'm also not 100% sure why the tabIndex is set - write a comment?

js/cards.js
```diff
@@ -114,6 +151,47 @@ export async function defineCardComponent() {
...
+    cardList[i].addEventListener("mouseover", addAnimation);
+    cardList[i].addEventListener("click", addAnimation);
+  }
+}
+export function removeHover(){
+  let cardList = document.getElementsByClassName("clicked");
+  let el;
+  for (let i = 0; i < cardList.length; i++) {
+    let z = window.getComputedStyle(cardList[i]).zIndex;
+    el = cardList[i].children[0].children[0].children[0];
```
miffens: what's the reason for the children[0] chain instead of a query that's more readable?

js/cards.js
```diff
@@ -114,6 +151,47 @@ export async function defineCardComponent() {
...
+    if(parseInt(z) !== card_count){
```
miffens: I don't fully understand these lines - I'd expect these event listeners would have to be recalculated each time a card is flipped, but I only see `removeHover` called once in main.js

js/main.js
```diff
@@ -71,7 +72,26 @@ async function main() {
   getCSVData(async (cards_data) => {
     switch (CURRENT_PAGE) {
       case "/gacha.html":
-        GACHA_BUTTON.onclick = (event) =>
+        GACHA_DISPLAY.forEach((elem) => {
```
miffens: 1. Perhaps the name GACHA_VIEW_SETTING would be a clearer name. Gacha display made me think this was the whole gacha game container
2. The `forEach` confused me - document.querySelectorAll('input[name="gacha-display"]')[0] so that you don't need to loop over 1 item here would be more understandable to read

js/main.js
```diff
@@ -71,7 +72,26 @@ async function main() {
...
+              GACHA_SECTION.classList.add("pile-display");
+              removeHover();
+            }
+          });
+        });
+        if(window.innerWidth <= 800){
+          document.getElementById("gacha-grid").click();
+        } else {
+          document.getElementById("gacha-pile").click();
+        }
```
miffens: I don't think we need this, unless Bing specifically requested it.

If you want to keep it, is it possible to set the value of the input directly instead of doing it through a "click"? This can be addressed later, just let me know and we'll add it to the Notion

miffens (commented):
Rest of comments for now. I'm not fully understanding the full lifecycle of how the animation/ hover effects are added/ removed as cards get flipped and will have another stab at it after a full night's sleep :')

js/dialog.js
miffens: Added task to Notion for followup later - no changes needed here for this PR

js/dialog.js
miffens: Added a Notion task to follow up on this later if we have time - optional to address within this PR if the existing code is functional

css/card.css
kirinarimasu: Yep, I will remove the stuff I added. :)

css/card.css
kirinarimasu: Updated this to .card-subtitle in line with o5v3's updates - keeping only the text colour CSS.

css/card.css
miffens: Added task to Notion

css/card.css
kirinarimasu: There are currently separate animations set up for the flip and the moving across (when in the 'pile' display format). I added some comments to try and explain the setup.

css/card.css
kirinarimasu: I'll remove it, it's not a necessity. :)

css/card.css
kirinarimasu: This has been updated to align to o5v3's updates.

js/cards.js
kirinarimasu: Missed this while scrolling, thank you for the pickup!

js/cards.js
kirinarimasu: This was a leftover bit from the previous code - have deleted this too now. :)

js/cards.js
kirinarimasu: Update this name for clarity. Originally it was literally used for counting how many cards had been opened, haha. :)

js/cards.js
kirinarimasu: Makes perfect sense, d'oh. :) Have updated this as suggested.

js/cards.js
kirinarimasu: Deleted that console.log! Might change this to what bon suggested with the 'unclickable' class.

css/card.css
kirinarimasu: Ah the @media-screen further down is actually the culprit here, I don't think going with pure vh is actually the best way to go with this. Made a small adjustment here, but open to suggestions.

js/cards.js
kirinarimasu: The particular code here is no longer present, but I have added some comments to explain what is happening.

js/main.js
kirinarimasu: I updated the name, and refactored the code a bit to remove the forEach. Hopefully it's a little easier to read!

js/cards.js
kirinarimasu: The tabIndex is set to -1 so that if you try to use keyboard controls to flip the card, you don't have to tab twice on each card (as it stops on both card-front and card-back, but I only want card-back to be stopped on). If you stop on card-front and hit enter there's no action. If you stop on card-back, it will flip the card.

kirinarimasu: Updated the name to `setupCardGacha`, but easily updated to something else if that makes more sense.

js/cards.js
kirinarimasu: I'm not sure why this seemed easier than using a different class for example - I've updated this so hopefully it reads a bit better!

js/cards.js
kirinarimasu: There was actually some duplicated code, so this is a good call - I have reused removeHover instead and redone the code so that it hopefully makes more sense!

js/cards.js
kirinarimasu: This _might_ be better with the new setup. Please let me know if you can see it still lagging!

js/cards.js
kirinarimasu: Have updated this to use the unclickable class, works nicely. :)

js/dialog.js
kirinarimasu: Thanks, it didn't actually make a whole lot of sense until I took a closer look at what was happening there. Makes sense to include it there, and the CSS bon suggested really helps with this setup.

js/main.js
kirinarimasu: This makes sense to me to be different on mobile devices since I don't feel like the pile layout is ideal with that. I have updated the way this is done though to not use a click, and instead to check it with more typical JavaScript and then just dispatch a new change event.

js/dialog.js
kirinarimasu: This is updated! :) Thank you to @bonbombs for that CSS suggestion.

js/dialog.js
kirinarimasu: I know you said to follow up later, but I have figured this out and updated. :)

js/cards.js
miffens: hmm nope, the flip + shine/rotate lags still. The hover to reanimate the rotation in both the gacha and dialog has been smooth though

js/cards.js
```diff
@@ -66,13 +71,16 @@ export async function defineCardComponent() {
...
-      image.src = this.getImageURL();
+      this.info = this.getElementsByClassName("card-info")[0];
+      this.image = this.getElementsByClassName("card-image")[0];
+      this.image.style.backgroundImage = 'url("' + this.getImageURL() + '")';
+      this.image.classList.add(setCardRarity(this.data["Rarity Folder"]));
       this.setupOnClickEvents();
       if (CARD_ART_HIDDEN_ON_LOAD) {
-        this.flipCard();
+        this.setupCardGacha();
       }
```
miffens: Main comment 1: My understanding is `defineCardComponent` runs once per card, but `setupCardGacha` runs on all 10 cards. Isn't that a lot of duplicate setup happening? Shouldn't `setupCardGacha` only run on the one card that is currently being defined?

js/cards.js
```diff
@@ -48,6 +48,11 @@ const sort_functions = {
...
     },
 };
+const HIGHEST_Z_INDEX = 10;
+let card_z_index = HIGHEST_Z_INDEX;
+let previous_card_front;
```
miffens: previous_card_front appears to no longer be used anymore

js/cards.js
```diff
@@ -85,15 +93,45 @@ export async function defineCardComponent() {
...
+
+      // register this as a clicked card, and add the class to animate to flip over
+      this.classList.add("clicked");
+      this.holder.classList.add("flip");
+
+      // Start the animation and update the z-index when the animation starts
+      // This is to stack any opened cards in the reverse order
+      this.addEventListener("animationstart", updateZIndex);
+      this.image.classList.add("opened");
+      this.image.classList.add("animated");
```
miffens, line 117: Main comment 2: Changing this to
```
      setTimeout(() => {
        this.image.classList.add("opened");
        this.image.classList.add("animated");
        // remove the previous hover effects only if the gacha is not displayed as a grid
        if(gacha_display_selection !== "gacha-grid"){
          this.addEventListener("animationend", removeHover);
        }
      }, 500);
```

Where 500 is the time it takes for the "flip" to finish (0.5s) fixed the flip lag issues for me. I guess trying to both flip and rotate was what was causing the lag

js/cards.js
```diff
@@ -114,6 +152,56 @@ export async function defineCardComponent() {
...
+  let cardList = document.getElementsByClassName("card-image");
+  for (let i = 0; i < cardList.length; i++) {
+    cardList[i].addEventListener("animationend", function(){
+      cardList[i].classList.remove("animated");
+    })
+    cardList[i].addEventListener("mouseover", addAnimation);
+    cardList[i].addEventListener("click", addAnimation);
+  }
+}
+export function addHover(){
```
miffens: to distinguish this better from `setupHover`, maybe name to `removeUnclickableFromCards`?

js/cards.js
```diff
@@ -114,6 +152,56 @@ export async function defineCardComponent() {
...
+  // remove any unclickable classes from all the opened cards, so all can be interacted with
+  let clickedCard = document.getElementsByClassName("opened");
+  for (let i = 0; i < clickedCard.length; i++) {
+    clickedCard[i].classList.remove("unclickable");
+  }
+}
+export function removeHover(){
```
miffens: `onlyHoverTopmostCard` or `transitionToPileHover` would be a clearer name (though a mouthful)

And to double check my understanding, this runs once on transition from grid to pile. Then, `flipCard` handles subsequent hover removal on the previously flipped card, and adds it to the one that was just flipped? Might be good to comment that

js/cards.js
```diff
@@ -114,6 +152,56 @@ export async function defineCardComponent() {
...
+  // check the current format is not grid
+  if(gacha_display_selection !== "gacha-grid"){
+
+    // this is the parent card container
+    let tcgCard = document.getElementsByClassName("clicked");
```
miffens: rename to `allFlippedCards` - `tcgCard` suggests it's only one card and it's unclear what state it's in

js/cards.js
```diff
@@ -114,6 +152,56 @@ export async function defineCardComponent() {
...
+    // this is the specific child the animation is applied to
+    let clickedCard = document.getElementsByClassName("opened");
+    // apply this to all the clicked cards, except the one with the z-index matching the current recorded z-index value
```
miffens, line 204: maybe expand this comment a bit, like:
// the one with the matching z-index is the most recently flipped card at the top of the pile, while the unmatching ones are beneath the most recently flipped card

js/cards.js
```diff
@@ -85,15 +93,45 @@ export async function defineCardComponent() {
       return `${CLOUDINARY_URL}/q_auto/${this.data["Rarity Folder"]}/${this.data["Filename"]}.png`;
     }
 
+    setupCardGacha() {
+      // Reset the initial z-index back to max pull
```
miffens: thank you for the added comments throughout!

js/dialog.js
```diff
@@ -13,9 +13,24 @@ export async function setupDetailsDialog() {
...
+  let dialogCard = DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0];
+  dialogCard.addEventListener("animationend", function() {
+    dialogCard.classList.remove("animated", "unclickable");
+  }, false);
+  dialogCard.addEventListener("mouseover", animateCard);
+  dialogCard.addEventListener("click", animateCard);
+
+  // animation reset
+  DETAILS_DIALOG_A11Y.on('hide', function (element, event) {
+    console.log("ayo?");
```
miffens: I laughed out loud, peak tempura coding mood - delete console.log

js/main.js
```diff
@@ -66,12 +67,37 @@ function toggleCollection(event) {
...
+    GACHA_SECTION.classList.add("grid-display");
+    GACHA_SECTION.classList.remove("pile-display");
+    addHover();
+  } else {
+    // make this a pile
+    GACHA_SECTION.classList.remove("grid-display");
+    GACHA_SECTION.classList.add("pile-display");
+    removeHover();
+  }
+}
```
miffens: (nit) this function be moved to cards.js since it's importing a bunch of things from it anyways. Then only one import is needed.
I don't feel strongly about this, feel free to keep as-is

miffens (commented):
2 main comments, rest are nits/ clearer code readability

js/cards.js
kirinarimasu: This has been removed.

js/cards.js
kirinarimasu: I wonder if it's better to add the delay in the CSS or by delaying the adding of the class with JavaScript?

js/cards.js
kirinarimasu: I've updated both `addHover` to `removeUnclickableFromCards`, and `removeHover` to `addUnclickableToCards` with a comment.

js/cards.js
kirinarimasu: Oh whoops you mentioned it here too. Updated it again and extra comment added. I don't really care if the name is long, if it makes sense to you. (It made sense to me but of course not for everyone. :))

js/cards.js
kirinarimasu: Sure. The reason I called it that is because of the name of the element itself, since there's already `clickedCards`, having both `openedCards` and `clickedCards` didn't make a lot of sense at the time. I've updated this.

js/cards.js
kirinarimasu: Updated this, with further clarifications.

js/dialog.js
kirinarimasu: AAAA I thought I deleted that! Maybe I did an 'undo' and forgot to remove it. I have removed it now.

js/cards.js
kirinarimasu: I've done it CSS way for now, because setTimeout I am sometimes wary of (it's only 500ms though so it's not a huge deal). We can swap it over if it makes no difference.

js/cards.js
kirinarimasu: Sorry this isn't a problem with the function being run again necessarily, just that there are a couple of things it should be doing only once... I've moved those to `renderCard` instead. The other two things remaining in that function probably could just be copied in, but originally it was a separate function in that if statement so I left it as-is for now.

js/main.js
kirinarimasu: I have updated this and moved it to cards.js.

js/cards.js
miffens: I don't like setTimeout either and I didn't know the CSS way, thanks for addressing!

js/dialog.js
```diff
@@ -48,3 +66,30 @@ export function updateDetailsDialog(data, cardUrl) {
...
 }
+
+function animateCard(dialogCard){
+  DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0].classList.add("animated", "unclickable");
+}
+function resetCardEffects(){
+  DETAILS_DIALOG_EL.getElementsByClassName("details-dialog-card")[0].classList.remove("holo", "basic", "secret", "rare", "ultra", "animated", "unclickable");
+}
+
+export function setCardRarity(folder){
```
miffens: (nit) purely a code organization thing - setCardRarity makes more sense to be placed in cards.js
optional/ can be addressed later

js/main.js
```diff
@@ -71,7 +71,19 @@ async function main() {
...
+        // this checks the box and dispatches a change event
+        if(window.innerWidth <= 800){
+          const gridInput = document.getElementById("gacha-grid");
+          gridInput.checked = true;
+          gridInput.dispatchEvent(new Event('change'));
+        }
+        
+        GACHA_BUTTON.onclick = (event) => 
           pullAndRenderCards(cards_data, COLLECTIONS_MAIN_CONTENT);
         await setupDetailsDialog();
```
miffens: delete setupDetailsDialog from gacha case now that it no longer opens the dialog

miffens (approved):

js/main.js
kirinarimasu: Thank you - removed this.

js/dialog.js
kirinarimasu: Makes sense! I have updated this. :)

PR #20, opened 2023-01-02
Setup pagination and choosing sort type

o5v3:
- Sets up a pagination system that breaks down the collection into many pages if the amount of cards to display is bigger than CARDS_PER_PAGE.
- Also implements some functions to sort cards, which can also sort in reverse order thanks to the use of closures.

That's basically it. I had made the "collection-pagination" container a nav element earlier and it looked pretty neat but one of the previous PRs must've done some css changes that make it impossible to see as a nav, so I changed it to a div. As a result the pagination controls are pretty basic (They literally have the default styling and are in the lower left corner), but I want to get the functionality out there first before focusing on anything else.


js/cards.js
```diff
@@ -107,32 +127,65 @@ function getOwnedCards(cards_data) {
...
       htmlLocation.innerHTML =
         "You have no cards at the moment. Try pulling some at the gacha!";
+      PAGINATION.container.classList.add("hidden");
+      return;
+    } else {
+      cards = sortCards(cards, sort, reverse);
+    }
+  }
+
+  if (cards.length > CARDS_PER_PAGE) {
```
miffens, line 173: (nit) comment explaining this section determines which page buttons to show/ hide?

maybe renaming PAGINATION to PAGINATION_BTNS would help make it clearer too

pages/collection.html
```diff
@@ -78,8 +78,22 @@ <h1>The Regis Altare Birthday Project 2023</h1>
...
+                <option value="Rarity-">Rarity (Reverse)</option>
+            </select>
+        </label>
+
         <p id="collected-cards-number">Collected cards:</p>
         <section id="card-list"></section>
+        <div id="collection-pagination">
+          <a id="pagination-previous" href="javascript:void(0)" class="hidden"><</a>
+          <a id="pagination-current" href="javascript:void(0)">1</a>
+          <a id="pagination-next" href="javascript:void(0)">></a>
```
miffens: For accessibility, I think button elements w/ aria-labels on the "<" and ">" is preferred. Or you can change them to "Next/ Previous" text

js/cards.js
```diff
@@ -107,32 +127,65 @@ function getOwnedCards(cards_data) {
...
   let ownedCount = 0;
   for (let rarity of RARITIES) {
     ownedCount += parseInt(localStorage.getItem(`count-${rarity}`) ?? "0");
   }
   COLLECTED_CARDS_NUMBER.textContent = `Collected cards: ${ownedCount}/${cards_data.length}`;
 
+  if (sort[sort.length - 1] === "-") {
+    sort = sort.slice(0, -1)
+    reverse = true;
+  }
```
miffens, line 155: Having trouble understanding this line. If 'sort' is one of the dropdown string values like "Rarity", when would it have "-"?

pages/collection.html
```diff
@@ -78,8 +78,22 @@ <h1>The Regis Altare Birthday Project 2023</h1>
...
+                <option value="Rarity">Rarity</option>
+                <option value="Rarity-">Rarity (Reverse)</option>
+            </select>
+        </label>
+
         <p id="collected-cards-number">Collected cards:</p>
         <section id="card-list"></section>
+        <div id="collection-pagination">
+          <a id="pagination-previous" href="javascript:void(0)" class="hidden"><</a>
+          <a id="pagination-current" href="javascript:void(0)">1</a>
```
miffens, line 95: As a quick change suggestion, could we also show total pages, like "1 of 2"? The "of 2" can be plain text

As a bigger change I'm not sure we need to do, I'm thinking about how important it is to have the functionality to quickly find specific cards without having to click 'next page' a bunch of times. I wonder if it'd be possible to have a "cards per page" dropdown with (10, 20, 50, show all) or be able to jump to specific pages via url (collection.html/5)

js/cards.js
```diff
@@ -107,32 +127,65 @@ function getOwnedCards(cards_data) {
...
-  cards.sort(sort_functions[sortType]);
+  let before = reverse ? -1 : 1
+  let after = reverse ? 1 : -1
+  cards.sort(sort_functions[sortType](before, after));
   return cards;
 }
 
-export function showCollection(cards_data, htmlLocation) {
-  let sort = localStorage.getItem("sort");
+export function showCollection(cards_data, htmlLocation, page = 1) {
```
miffens, line 139: this function is getting large, and the pagination button logic feels vulnerable to off by 1 bugs, but it works and I don't have any immediate suggestions to simplify things

pages/collection.html
```diff
@@ -78,8 +78,22 @@ <h1>The Regis Altare Birthday Project 2023</h1>
       <section>
         <button id="full-collection-toggle">Show Full Collection</button>
         <button id="reset-collection">Reset Card Collection</button>
+        <label>
+            Sort by:
```
miffens, line 82: I think this addition is causing a white border around the collection area and the rounded corners that were there before got lost. Could that be addressed in this PR or in a quick followup?

miffens (approved):

js/cards.js
o5v3: Done o7

pages/collection.html
o5v3: Changed them to "Next/Previous" text, since it does make more sense than "<>".

js/cards.js
o5v3: Added comment. If the sort type has a "-" at the end, it means it's in reverse order ("Rarity (Reverse)" in the dropdown has value "Rarity-", for example). It is admittedly a bit hacky, though effective. I can change it if it's _too_ hacky.

pages/collection.html
o5v3: Quick change done. o7

The cards per page dropdown can be done. The url part is somewhat more complex to ensure they can't get into invalid pages and stuff, but it can be done too. I'll work on that next PR.

js/cards.js
o5v3: Glad to see I'm not the only one who saw it was getting too big. Time to make a separate js file for the collection! I hope it doesn't slow down development though, I'll try to get it done ASAP.

As for the button logic I do think working with 1-based indexing is weird and usually leads to trouble, but imo it works better in this case as it wouldn't make sense for a user to be in page 0. I can add more safeguards around stuff just to be safer, I'll take note of that.

pages/collection.html
o5v3: The white border is caused by the section element that begins on line 78 and ends on line 98. I marked it with the "collection-border" id. It wraps the "card-list" section which is the one on blue and it's needed for the controls as everything in that section is replaced on card render. I did notice that deleting the "collection-border" doesn't bring back the round corners, so something else must be happening. I'll investigate it.

js/cards.js
miffens: no changes needed, thanks for the comment! I didn't pay attention to the value="Rarity-" in the HTML before

pages/collection.html
miffens: Actually I think the round borders is no longer there after Bon finalized the "tabs" page styling - now it's sharp borders with colored backgrounds. So it's an issue that the background is white, but the corners don't need to be round. I've added a ticket to address this on our Notion board

PR #21, opened 2023-01-03
Collection Dialog CSS Updates

bonbombs:


bonbombs:
@miffens ready to review for now but I anticipate I'll need to do additional tweaks & testing once https://github.com/elysiumdelivery/altare-tcg/pull/19 is merged.

css/card.css
```diff
@@ -14,21 +14,30 @@ tcg-card, .card-component, .card-front,
...
+  align-items: center;
+  justify-content: center;
+  row-gap: 2em;
+  column-gap: 1vmin;
 }
 #gacha-controls tcg-card {
   position: absolute;
   right:10%;
-  transition: all 0.3s ease; /* make any changes to this card flow, so it doesn't snap */
+  transition: transform 0.3s ease; /* make any changes to this card flow, so it doesn't snap */
```
bonbombs, line 40: explicit definition to reduce browser load on possibly transitioning all css properties associated with this element

css/card.css
```diff
@@ -399,10 +388,13 @@ section#gacha-controls input {
...
 }
 #gacha-controls .card-image {
+  position: absolute;
+  top: 0;
+  left: 0;
   box-shadow: 0 10px 20px -5px rgba(0, 0, 0, 0.5);
 }
 #gacha-controls .card-image, #gacha-controls .card-image:before, #gacha-controls .card-image:after {
-  transition: all 0.5s ease;
+  transition: transform 0.5s ease, filter 0.5s ease;
```
bonbombs, line 397: explicit definition to reduce browser load on possibly transitioning all css properties associated with this element

css/card.css
```diff
@@ -388,9 +378,8 @@ section#gacha-controls input {
   z-index: 10;
   touch-action: none;
   border-radius: 5%/3.5%;
-  box-shadow: -7px -7px 10px -5px transparent, 7px 7px 10px -5px transparent, 0 0 5px 0px rgba(255, 255, 255, 0), 0 55px 35px -20px rgba(0, 0, 0, 0.5);
-  transition: transform 0.5s ease, box-shadow 0.2s ease;
-  will-change: transform, filter;
+  box-shadow: 0 55px 35px -10px rgb(0 0 0 / 50%);
```
bonbombs, line 381: simplifying box-shadow for performance reasons

css/card.css
```diff
@@ -486,7 +478,7 @@ section#gacha-controls input {
   background-blend-mode: overlay;
   z-index: 2;
   filter: brightness(1) contrast(1);
-  transition: all 0.33s ease;
+  transition: filter 0.33s ease;
```
bonbombs, line 481: explicit definition to reduce browser load on possibly transitioning all css properties associated with this element

css/main.css
```diff
@@ -52,26 +52,40 @@ footer svg {
...
+button:not(.card-front):not(.card-back):hover,
+select:hover {
   background-color: var(--accent-color);
 }
 
+#size-dropdown {
+  width: auto;
+}
+
+#page-header {
```
miffens, line 81: While you're here, want to make the "The Regis Altare Birthday Project 2023" header text white as well?

js/main.js
```diff
@@ -82,6 +82,12 @@ async function main() {
           gridInput.checked = true;
           gridInput.dispatchEvent(new Event('change'));
         }
+        // Otherwise, default to pile and assign pile-display class by default
+        else {
+          const pileInput = document.getElementById("gacha-pile");
+          pileInput.checked = true;
+          pileInput.dispatchEvent(new Event('change'));
+        }
```
miffens, line 90: does this need to be explicitly written? I assumed this would happen by default - what happens if we don't have this?

miffens (approved):
idk what you did exactly to address the pixel gap between the tab and main area but the preview looks good to me :D

js/main.js
bonbombs: Yeah. w/o this, the radio button would get selected visually but the change event that assigns the `pile-display` class to the appropriate container doesn't fire, which causes some weirdness when I tried to scale the page to portrait. I believe there are some specific CSS rules that looks for `.pile-display`.

PR #22, opened 2023-01-03
Add page size dropdown and card titles in collection page.

o5v3:
Fairly small PR (ignoring Prettier doing its thing):

- Creates a page size dropdown for showing 10, 20, 50 or all cards per page.
- Shows card number and title below each card in the collection page.
- Adds Collector Number (Reverse) to sort dropdown. (Only 1 line of code)
- Changed pagination controls from "a" to "button", and moved it all into 1 line.

I wanted to get this out of the way to focus on making collections.js next.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -12,7 +12,8 @@ const FULL_COLLECTION_TOGGLE = document.getElementById(
   "full-collection-toggle"
 );
 const RESET_COLLECTION = document.getElementById("reset-collection");
-const SORT_DROPDOWN = document.getElementById("sort-dropdown")
+const SORT_DROPDOWN = document.getElementById("sort-dropdown");
+const CARDS_PER_PAGE_DROPDOWN = document.getElementById("size-dropdown");
 
 export const CARD_ART_HIDDEN_ON_LOAD =
   PAGES_WHERE_CARD_HIDDEN.includes(CURRENT_PAGE);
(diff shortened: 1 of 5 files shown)


pages/collection.html
```diff
@@ -82,18 +82,27 @@ <h1>The Regis Altare Birthday Project 2023</h1>
...
         <p id="collected-cards-number">Collected cards:</p>
         <section id="card-list"></section>
-        <div id="collection-pagination">
-          <p id="pagination-total"></p>
-          <a id="pagination-previous" href="javascript:void(0)" class="hidden">Previous</a>
-          <a id="pagination-current" href="javascript:void(0)">1</a>
-          <a id="pagination-next" href="javascript:void(0)">Next</a>
+        <div id="collection-pagination" class="hidden">
+          <button id="pagination-previous" class="hidden">Previous</button>
+          <button id="pagination-current">1</button>
```
miffens: the pagination current text doesn't need to be a button :)

js/main.js
```diff
@@ -85,14 +86,22 @@ async function main() {
         RESET_COLLECTION.onclick = (event) => {
           localStorage.clear();
           localStorage.setItem("fullCollection", "true");
-          localStorage.setItem("sort", SORT_DROPDOWN.value)
+          localStorage.setItem("sort", SORT_DROPDOWN.value);
+          localStorage.setItem("page-size", CARDS_PER_PAGE_DROPDOWN.value);
```
miffens, line 90: do these need to be set when clicking the reset button? What happens if they aren't?

miffens (approved):

pages/collection.html
o5v3: oops. Changed to span. o7

js/main.js
o5v3: Functionally, nothing, :D
They're there to remember the dropdown values after the reset, If they aren't there, everything will work but if you had "Sort by Rarity" set ,for example, then if you look at the full collection right away it would be sorted with the default, which is collector number and not the one shown in the dropdown. It's to cover a corner case basically.

PR #25, opened 2023-01-07
Gacha accessibility

miffens:
- Moved button down a bit closer to the card pile section
 - Added rarity to card subtitle
 - Show subtitle on flipped cards: for pile, show it at the top (showing it under the card would slant it in the direct of the card). for grid, show directly under card
 - Screenreader compatibility for gacha game

This video shows:
 - The card grid recommendation note, and me setting it to card grid
 - Flipping over cards
 - re-focusing on already flipped cards
 - the "no more cards" message if you reach the end of the card list

https://user-images.githubusercontent.com/47371080/211174180-27f1c36e-8c36-41f9-bdd4-73ad5b489bc8.mov

### DIFF
--- a/css/card.css
+++ b/css/card.css
@@ -43,8 +43,13 @@ tcg-card {
   right: 10%;
   transition: transform 0.3s ease; /* make any changes to this card flow, so it doesn't snap */
 }
-tcg-card .card-subtitle {
+.card-subtitle {
   color: white;
+  user-select: text;
+}
+#pile-pull-announcement {
+  margin: 0;
+  text-align: center;
 }
 
 /* rotate card differently depending on its position in the pile */
@@ -221,10 +226,6 @@ tcg-card .card-subtitle {
   right: 1em;
 }
 
-.card-subtitle {
-  user-select: text;
-}
-
 #sort-dropdown {
   text-align: center;
   margin-left: 4px;
@@ -235,7 +236,7 @@ section label {
 }
 
 #gacha-button {
-  margin-top: 1em;
+  margin-bottom: 1em;
   background-color: var(--accent-color);
 }
 
--- a/css/card.css
+++ b/css/card.css
@@ -109,13 +109,13 @@ tcg-card {
   animation: left 1s ease 0s 1;
   animation-fill-mode: forwards;
 }
-#gacha-controls tcg-card.clicked:nth-child(3n + 1) {
+#gacha-controls.pile-display tcg-card.clicked:nth-child(3n + 1) {
   --rotation: -3deg;
 }
-#gacha-controls tcg-card.clicked:nth-child(3n + 2) {
+#gacha-controls.pile-display tcg-card.clicked:nth-child(3n + 2) {
   --rotation: 5deg;
 }
-#gacha-controls tcg-card.clicked:nth-child(3n + 3) {
+#gacha-controls.pile-display tcg-card.clicked:nth-child(3n + 3) {
   --rotation: 0deg;
 }
 
--- a/css/card.css
+++ b/css/card.css
@@ -47,7 +47,8 @@ tcg-card {
   color: white;
   user-select: text;
 }
-#pile-pull-announcement {
+#pull-announcement {
+  color: white;
   margin: 0;
   text-align: center;
 }
--- a/js/cards.js
+++ b/js/cards.js
@@ -32,16 +32,15 @@ export async function defineCardComponent() {
       this.holder = this.getElementsByClassName("card-component")[0];
       this.front = this.getElementsByClassName("card-front")[0];
       this.back = this.getElementsByClassName("card-back")[0];
-      this.info = this.getElementsByClassName("card-info")[0];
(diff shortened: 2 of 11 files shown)


css/card.css
```diff
@@ -43,8 +43,13 @@ tcg-card {
   right: 10%;
   transition: transform 0.3s ease; /* make any changes to this card flow, so it doesn't snap */
 }
-tcg-card .card-subtitle {
+.card-subtitle {
```
miffens, line 46: Made more general since using card-subtitle class for pile-pull-announcement at the top of the pile card view

css/card.css
```diff
@@ -235,7 +236,7 @@ section label {
 }
 
 #gacha-button {
```
miffens, line 239: Moved button below the gacha notes

js/cards.js
```diff
@@ -79,6 +84,12 @@ export async function defineCardComponent() {
       // remove the previous hover effects only if the gacha is not displayed as a grid
       if (gacha_display_selection !== "gacha-grid") {
         this.addEventListener("animationend", addUnclickableToCardsExceptLast);
+        document.getElementById("pile-pull-announcement").textContent =
+          this.getSubtitleText();
+      } else {
+        this.subtitle.classList.remove("hidden");
+        // Set aria-live on parent container to announce subtitle text when it appears
+        this.setAttribute("aria-live", "polite");
```
miffens: todo: since I set aria-hidden on the subtitle, this live announcement doesn't work anymore
I'll probably have pile-pull-announcement in both gacha views and have it be visually hidden in grid view

js/cards.js
```diff
@@ -96,13 +107,17 @@ export async function defineCardComponent() {
...
-      this.front.insertAdjacentHTML(
+    getSubtitleText() {
+      return `${this.data["Collector Number"].padStart(3, "0")} ${
+        this.data["Card Display Name"]
+      } (${this.data["Rarity Folder"]})`;
+    }
+
+    addCardSubtitle() {
+      // aria-hidden since the card has its own aria-label handling
+      this.insertAdjacentHTML(
```
miffens: linking the subtitle to the image with aria-labelledby is usually best practice, but our card html structure is unconventional compared to generic websites. Manipulating aria-label was the quickest way I could achieve conditional labelling

js/main.js
```diff
@@ -119,7 +119,7 @@ async function main() {
         }
 
         GACHA_BUTTON.onclick = (event) =>
-          pullAndRenderCards(cards_data, COLLECTIONS_MAIN_CONTENT);
```
miffens, line 122: while I was here: delete unused arg

pages/gacha.html
```diff
@@ -145,14 +145,17 @@ <h1 id="title-header">The Regis Altare Birthday Project 2023</h1>
           />
           <label for="gacha-grid"> Card Grid</label>
         </form>
-        <button id="gacha-button" type="submit">Open Card Pack</button>
         <div id="gacha-notes">
+          <p class="visually-hidden">
+            Card Grid is the recommended view for screenreader users.
+          </p>
```
miffens, line 159: In pile view, the screenreader loses its position if you flip cards "out of order":
 - you're in grid view and go down to flip the 7th card, then you switch back to pile view. I don't expect screenreaders to use the site in this way since the view is purely visual, so I'm not concerned about solving this
 - you can reach the lower cards with screenreader and flip them), which disorients screenreader users. The solution would be to programmatically toggle tabbing and aria-hidden as cards are flipped so only the frontmost card is accessible, but that would be a nontrivial effort. So I compromised with this note for screenreader uses :')

js/cards.js
miffens: done

js/cards.js
miffens: moved card-subtitle to card.html and commented this in the file

css/card.css
miffens: outdated comment, but I think it's fine to keep the removal of tcg-card

css/card.css
```diff
@@ -104,13 +110,13 @@ tcg-card .card-subtitle {
   animation: left 1s ease 0s 1;
   animation-fill-mode: forwards;
 }
-#gacha-controls tcg-card.clicked:nth-child(3n + 1) {
+#gacha-controls.pile-display tcg-card.clicked:nth-child(3n + 1) {
```
miffens, line 113: Without this increased specificity, the subtitle text animates on rare+ cards in grid view flip

pages/artist-writer-board.html
```diff
@@ -41,7 +41,9 @@
     <main>
       <h1>The Regis Altare Birthday Project 2023</h1>
 
-      <a href="#artist-writer-board" class="visually-hidden">Skip to content</a>
+      <a href="#artist-writer-board" class="visually-hidden skip-link"
+        >Skip to content</a
+      >
```
miffens, line 46: added white color. the default blue is difficult to see

bonbombs (approved):
Looks and works great!!

PR #27, opened 2023-01-10
Credits Page

bonbombs:
Adding contributor & credits csvs, as well as adding in the chibi Altare to the home nav. Also additional small tweaks to some css styles for responsive views


js/messages.js
```diff
@@ -1,46 +1,115 @@
...
-
-function getCSVData(callback = undefined) {
-    Papa.parse(CSV_FILENAME, {
-      download: true,
-      //To treat the first row as column titles
-      header: true,
-      complete: (result) => {
-        if (callback) {
-          callback(result);
+const MESSAGES_CSV = "../contributor_data.csv";
```
bonbombs, line 1: @miffens Currently the code to pull csv data for the messages and credits are in a separate file. Not sure if that's too redundant and if it should just go into main.js instead.

css/main.css
```diff
@@ -22,6 +23,7 @@ html {
 body,
 main {
   background: transparent;
+  margin: 8px;
```
bonbombs, line 27: adding a bit of extra spacing, mostly for aesthetic reasons. Mobile view looked especially tight.

css/main.css
```diff
@@ -89,6 +91,8 @@ select:focus-visible {
 #title-header {
   color: white;
   text-align: center;
+  margin-top: 1em;
```
bonbombs, line 109: Reducing spacing here on the title header

css/main.css
```diff
@@ -99,11 +103,16 @@ select:focus-visible {
 #collection-pagination {
   display: flex;
   align-items: center;
-  justify-content: center;
+  justify-content: space-around;
```
bonbombs, line 138: More spacing. Makes it look better in portrait/narrow view.

css/main.css
miffens: would be helpful to comment this in the code so that the next person modifying knows what they'll affect if they change it

same for other comments

css/nav.css
```diff
@@ -37,6 +37,39 @@ nav a .icon {
...
 }
 
+/* chibi altare always sits on home tab */
+nav a#nav-home #chibi-altare {
+  content: "";
+  position: absolute;
+  /* background-position: 50% 100%; */
+  /* background-size: contain; */
+  /* background-repeat: no-repeat; */
+  /* background-image: url(../images/ChibiAltare.png); */
```
miffens: delete commented out code

js/messages.js
```diff
@@ -1,46 +1,115 @@
...
+}
+
+
+function renderMessages (message_data) {
+    let messageContainer = document.getElementById("artist-writer-messages");
+    for (var i = 0; i < message_data.length; i++) {
+        let message = message_data[i];
+        let messageNode = document.createElement("li");
+        // TODO: Dummy data for now. Alternate roles for now as a test. This should be pulled from a csv somewhere.
+        //   let userRole = i % 2 === 0 ? "Artist" : "Writer";
```
miffens: delete outdated comment - we have the real data now

js/messages.js
miffens: You could change `getCSVData` to take a filename param, then import that from main.js so you don't have to almost-copy it for messages.js. I don't feel strongly about this though. Keeping as-is is fine

pages/credits.html
```diff
@@ -136,7 +134,11 @@ <h1 id="title-header">The Regis Altare Birthday Project 2023</h1>
         </button>
     </nav>
 
-      <section id="credits"></section>
+      <section id="credits">
+
+        <h1>Project Credits</h1>
```
miffens: (nit) you can link the `h1` to `section` with aria-labelledby so that when a screenreader is browsing by section, the section will be read out like "Section: Project Credits" instead of just "Section"

So like
```
      <section aria-labelledby="credits-header" id="credits">

        <h1 id="credits-header">Project Credits</h1>
```

https://accessibility.psu.edu/screenreaders/voiceover/ (see 'Rotor Window' - it's a convenient function that lets users jump to specific headers/ sections, which is why it's important to use those HTML tags. It's something like ctrl + cmd + U on VoiceOver)

miffens (approved):

PR #28, opened 2023-01-10
Add search bar to collection page

o5v3:
Adds a search bar that will use the query provided to search for cards that include it in the following fields:

- Collector Number.
- Card Display Name.
- Artist Credit.
- Writer Credit.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -87,7 +87,7 @@ function setupCollectionControls() {
   }
   RESET_COLLECTION.onclick = (event) => {
     localStorage.clear();
-    localStorage.setItem("showFullCollection", "true");
+    localStorage.setItem("showFullCollection", "false");
     localStorage.setItem("sort", SORT_DROPDOWN.value);
     localStorage.setItem("page-size", CARDS_PER_PAGE_DROPDOWN.value);
     toggleCollection();
--- a/js/main.js
+++ b/js/main.js
@@ -93,7 +93,7 @@ function setupCollectionControls() {
     toggleCollection();
   };
 
-  SEARCH_BAR.oninput = (event) =>
+  SEARCH_BAR.onchange = (event) =>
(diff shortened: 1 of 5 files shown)


js/main.js
```diff
@@ -91,6 +92,10 @@ function setupCollectionControls() {
     localStorage.setItem("page-size", CARDS_PER_PAGE_DROPDOWN.value);
     toggleCollection();
   };
+
+  SEARCH_BAR.oninput = (event) =>
+    showCollection(cards_data, COLLECTIONS_MAIN_CONTENT, 1, event.target.value);
+
```
miffens: if it's not too much extra work, maybe add a debounce here?

pages/collection.html
```diff
@@ -147,6 +147,7 @@ <h1 id="title-header">The Regis Altare Birthday Project 2023</h1>
             <option value="all">All</option>
           </select>
         </label>
+        <input type="search" name="search" id="search-bar" placeholder="Search by name/artist/writer..." aria-label="Search cards by name, artist or writer" size="23" maxlength="32">
```
miffens: Any reasoning behind the max length choice?

miffens (commented):
Let me push some CSS for the search bar into your PR, then I'll approve

miffens (approved):

js/main.js
o5v3: Changed it from "oninput" to "onchange". oninput executes on every character while onchange executes when pressing enter or losing focus, so it executes only once per query without having to account for edge cases while debouncing.

pages/collection.html
o5v3: Letting the users type input without character limit could lead to some performance issues; I'm not certain on that, but regardless 32 characters should be more than enough to narrow down the search, so I placed the limit just in case.

PR #30, opened 2023-01-18
Loading screen removal. Added bouncing slime to gacha page

bonbombs:
also added additional optimizations to card animation


miffens (approved):
Did a bit of manual checking and skimmed code - changes make sense

Especially now that we're adding more animation to the site, I get a bit motion sick interacting with it and wish we had a reduced motion option on the site, which animation-heavy sites w/ accessibility do have. I thought about a quick `transform: none !important` toggle but that would probably make some transitions jank/ break card flip positioning. Curious if you have any ideas?

PR #31, opened 2023-01-21
Gacha Page - Reduce Motion

kirinarimasu:
Added extra JS and CSS to handle reducing the motion of the cards - mainly the card shake, while keeping the movement left/right and flip. This is only for the gacha page.


js/cards.js
```diff
@@ -198,6 +201,56 @@ export function updateGachaView(e) {
       .classList.remove("visually-hidden");
   }
 }
+export function updateGachaMotion(){
+  // get the list of cards
+  let cardImageList = document.getElementsByClassName("card-image");
+  let cardList = document.getElementsByTagName("tcg-card");
+  // document.getElementsByClassName("card-component");
```
miffens: delete commented out code

js/cards.js
```diff
@@ -254,6 +307,12 @@ export function renderCards(cards, htmlLocation, replace = false) {
   }
   // setup hover for all cards
   setupHover();
+
+  // only on the gacha page, but we want to do this after cards are added
+  if(document.getElementById("gacha-controls") !== null){
+    // update the motion settings based on the checkbox
+    updateGachaMotion();
+  }
```
miffens, line 314: Is this line strictly needed if the checkbox setting isn't saved in localStorage and always starts unchecked?

js/main.js
```diff
@@ -108,7 +110,8 @@ async function main() {
         // watch for any selection changes - either grid or pile card display
         GACHA_VIEW_SETTING[0].addEventListener("change", updateGachaView);
         GACHA_VIEW_SETTING[1].addEventListener("change", updateGachaView);
-
+        // add or remove the shine animation on the cards to reduce the amount of animations there are
```
miffens: (nit) it doesn't just control shine animation, so "add/ remove game animations" would be a more accurate comment. No comment at all would be OK too, the code is self-explanatory imo

css/card.css
```diff
@@ -301,6 +301,13 @@ section#gacha-controls label {
 section#gacha-controls input {
   margin-left: 15px;
 }
+section#gacha-controls fieldset:nth-child(2){
```
miffens, line 304: Do these CSS adjustments pertain to the animation changes? How, if so?

miffens (approved):

js/cards.js
kirinarimasu: Oops, there's always something! Removed this.

js/cards.js
kirinarimasu: If it's not there it doesn't apply the "remove-motion" class to the cards upon the cards being generated (for some reason, probably since the check box doesn't get selected again so it's not a 'change'.).

js/main.js
kirinarimasu: Sorry, the behaviour changed after our discussion today so I forgot to update this. I have removed it.

css/card.css
kirinarimasu: The line that contains `fieldset:nth-child(2)` specifically is there to style the checkbox and separate it a bit further from the Card Pile/Card Grid radio buttons.

PR #33, opened 2023-01-23
Collections Foreword and Collections Page Adjustments

kirinarimasu:
Added collections foreword and made some other styling adjustments to the buttons for smaller sized screens.

### DIFF
--- a/css/card.css
+++ b/css/card.css
@@ -268,8 +268,14 @@ tcg-card {
   margin-left: 4px;
 }
 
+#reset-collection {
+  margin-right: 15px;
+}
+
 section label {
-  padding-left: 15px;
+  padding-right: 15px;
+  display: inline-block;
+  white-space: nowrap;
 }
 .grid-display.pulled section {
   padding-bottom: 70px;
@@ -282,10 +288,6 @@ section label {
 }
 .pile-display button.gacha-button {
   margin-bottom:0.5em;
-}
-
-#search-bar {
-  margin-top: 15px;
 }
 
 #gacha-controls button.gacha-button:hover,
@@ -391,6 +393,12 @@ section#gacha-controls fieldset:nth-child(2) > input:first-child {
   }
 }
 
+@media screen and (max-width: 1320px) {
+  #search-bar {
+    margin-top: 15px;
+  }
+}
+
 @media screen and (max-width: 1000px) {
   #gacha-controls {
     /* make the card even smaller whtn the screen is resized further */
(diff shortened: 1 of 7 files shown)


css/card.css
```diff
@@ -284,10 +290,6 @@ section label {
   margin-bottom:0.5em;
 }
 
-#search-bar {
```
kirinarimasu, line 287: This was causing the entire section to be bumped down, so removed it here and added a media query instead to apply the same for smaller screens.

pages/collection.html
```diff
@@ -175,6 +175,12 @@ <h1 id="title-header">The Regis Altare Birthday Project 2023</h1>
           </label>
         </section>
 
+        <div id="collection-foreword" class="hidden">
+          <p><strong>Disclaimer:</strong> The cards in our lovingly fan-curated RegIsekai deck run on pure creativity, and are not regulated or balanced for play. The unplayability is an artistic choice to reflect the chaotic evil of the Demon God King Regis himself, and definitely not because balancing a playable TCG is really, really hard.</p>
+          <p>I mean… you can still <em>try</em> to play it, but a mald-free experience is far from guaranteed.</p>
+          <button id="foreword-close" aria-label="close" type="button">Dismiss</button>
```
miffens: aria-label isn't needed since the button already has the text Dismiss. aria-label is usually only used for icon buttons w/ no text

pages/message-board.html
```diff
@@ -137,16 +137,25 @@ <h1 id="title-header">The Regis Altare Birthday Project 2023</h1>
...
-        <video autoplay muted loop controls>
-          <source
-            src="https://res.cloudinary.com/dazcxdgiy/video/upload/v1669320701/kudoboard_qbu1gi.mp4"
-            type="video/mp4"
-          />
-          Your browser does not support the video tag.
-        </video>
+        
+        <!-- Link out to kudo board -->
+        <p>This preview is best viewed on desktop.
```
miffens, line 142: "This preview is best viewed on desktop." I assume is an old message from when we tried the giga JPEG. I think we can delete this sentence now?

miffens (approved):

pages/message-board.html
kirinarimasu: I added it because I considered that even with the video it was still best viewed on Desktop, because I don't expect a tiny video would be very legible to those on their phone.

pages/collection.html
kirinarimasu: Thanks, I can remove. I wasn't sure if Dismiss and Close would be considered the same thing to be honest, so I thought I'd just add it for clarity.

pages/message-board.html
miffens: good point, I'm good with keeping it then

PR #37, opened 2023-01-24
updated gacha cards

bonbombs:


css/welcome.css
```diff
@@ -323,6 +333,11 @@ main {
     font-size: 1.8vmin;
   }
 
+  /* hide the text and let the card imgs read out the alt text */
+  #welcome-nav a span {
+    display: none;
+  }
+
```
miffens: this may not be needed - can just delete the text span

css/welcome.css
```diff
@@ -313,6 +294,35 @@ main {
...
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+  }
+
+  #welcome-nav a.special-nav-card {
+    /* border gradient only seems to work when adding it to the bg property */
+    border-color: transparent;
+    background: linear-gradient(45deg, #a7daf5, #ffffff, #a7daf5) padding-box,
```
miffens: The glow behind the card seems to not be perfectly aligned with the new images

miffens (commented):
I might push changes into this PR to address my own comments since I know you're busy!

css/welcome.css
bonbombs: Removed in the lastest commit! No need for it now that we have the bg image

css/welcome.css
bonbombs: I think we still need this for the text to show in the mobile view. I can revise the above comment to clarify

miffens:
The card glow on hover still seems to not be evenly around the card? Or is this intentional :0
card glow

kirinarimasu:
> The card glow on hover still seems to not be evenly around the card? Or is this intentional :0 card glow

It looks to me like the card image is bigger than the anchor tag, so it's being clipped by the card image. 

Suggestion: Possible fix (bit hacky maybe), the card height is 1.395x larger than the width:

```
#welcome-nav a {
    /* adjust the width and height in welcome.css */
    width: 13vmin;
    height: calc(13vmin*1.3975);
    /* image corners need to be avoided too, could be come other value other than 5px */
    border-radius: 5px;
}
```

bonbombs:
> calc(13vmin*1.3975)

this definitely makes sense. I updated the code w/ those changes, thank you!!

css/welcome.css
```diff
@@ -94,35 +94,19 @@ main {
 #welcome-nav a {
   position: absolute;
   width: 13vmin;
-  height: 17vmin;
-  background: white;
-  border-radius: 5px;
-  border: 0.5vmin solid gray;
+  height: calc(13vmin*1.3975);
```
miffens, line 97: (nit) add a comment explaining this is a hack to get the a tag's size to be the same as the card image

miffens (approved):

PR #38, opened 2023-01-24
Add CSV File and make adjustments to make it work

o5v3:
Whew, this is a relatively big one. Don't let the formatting changes from Prettier fool you though, the code changes themselves are small. Turns out some things were dependent on Collector Number which broke the Element cards display, but it's nothing an special case couldn't solve. The big change is that now we're finally using the real cards! WOOOOO!

Oh, and changing from cloudinary to netlify is practically done. With the change to the source url it's only:
-Delete Netlify.toml plugin configuration.
-Deactivate plugin in Netlify dashboard.

And after that, our images will be served directly from Netlify!


o5v3:
Hmmm, there seems to be a problem with deterministic ordering of the Element cards. Every time they're rendered, they appear in a different order. I'll look into this.

js/gacha.js
```diff
@@ -8,74 +8,83 @@ export const GACHA_BUTTONS = document.getElementsByClassName("gacha-button");
...
+    UltraRare: 35,
+    SecretRare: 15,
   },
 };
 
 //Pulls a set of cards guided by an array of "slots" passed to the function.
 function pullCards(slots) {
   let pullCount = parseInt(localStorage.getItem("pull-count") ?? "0") + 1;
-  let pulledIDs = [];
+  //let pulledIDs = [];
```
miffens: delete commented code

js/gacha.js
```diff
@@ -112,8 +135,9 @@ function pullCards(slots) {
       }
     }
   }
-  localStorage.setItem("pull-count", pullCount);
-  localStorage.setItem(`pull-${pullCount}`, JSON.stringify(pulledIDs));
+  //Commenting because we aren't doing anything with this yet... if ever.
+  //localStorage.setItem("pull-count", pullCount);
+  //localStorage.setItem(`pull-${pullCount}`, JSON.stringify(pulledIDs));
```
miffens, line 144: no strong opinion about deleting/ keeping this though

js/gacha.js
```diff
@@ -86,9 +95,23 @@ function pullCards(slots) {
     //This applies to as many rarities as declared in the slot.
     for (let rarity in slot) {
       if (dice <= slot[rarity]) {
-        card = getRandomCards(cards_by_rarity[rarity], 1)[0];
+        pulled_cards = getRandomCards(cards_by_rarity[rarity], 4);
```
miffens, line 102: To check my understanding: for each card pull, pick 4 random cards. Then among those 4 cards, pick the first non-dupe.
In the slim chance occurs that all 4 pulled_cards are already in `cards`, the dupe will get added to the 10-pull? Maybe add a comment about that

js/cards.js
```diff
@@ -33,6 +36,10 @@ export async function defineCardComponent() {
       this.data = cards_data.find(
         (card) => card["Collector Number"] == this.getAttribute("card-id")
       );
+      if (this.data["Rarity Folder"] === "Element") {
+        this.data = Object.assign({}, this.data);
+        this.data["Collector Number"] = "000";
+      }
```
miffens: I assume this was added before the Elements had a collector number in the CSV. Is this still needed now that the latest CSV has collector number 000-1, 000-2 etc?

miffens (approved):

js/gacha.js
o5v3: Done o7

js/gacha.js
o5v3: Correct. The chance of the 4 cards being duplicates is extremely slim, virtually impossible, but if it ever happens it's better to return the duplicate rather than risk an infinite reroll situation.

js/cards.js
o5v3: This was added after that. By doing this, the system internally manages element cards with their assigned IDs but uses 000 for everything visually related. I added a comment; the Object.assign call copies the data object by value so it can be changed without it affecting its ID on the cards_data array. (since objects are passed by reference by default in JS)

PR #39, opened 2023-01-24
General Non-Landing Styling Updates

kirinarimasu:
Main change: add PDF / accessible view buttons to footer of all pages as well as at the top of the credits section
Misc:
 - Card animation issue for Firefox
 - Polish layout shift of pull announcement on gacha page

(miffens wrote this to keep track of all the open PRs better)


miffens:
Quick comments from just looking at the preview:

Think you missed the github icon/ link on the welcome page :)
The "made with love by altventurers" is a bit hard to see against certain background colors. Maybe add a dark stroke?
I think we should keep the github link at the footer

kirinarimasu:
@miffens Thanks! I didn't notice the welcome page at all /bonked. Made the changes you suggested!

pages/collection.html
```diff
@@ -189,19 +189,36 @@ <h2 id="collected-cards-number">Collected cards:</h2>
       </section>
 
       <footer>
-        <a href="https://github.com/elysiumdelivery/altare-tcg"
-          ><svg
-            role="img"
-            viewBox="0 0 24 24"
-            xmlns="http://www.w3.org/2000/svg"
+        <div id="footer-nav">
+          <a id="footer-download-pdf" class="download-button" href="#">
```
miffens, line 193: I assume the hrefs are blank intentionally since those aren't ready yet. I've made a separate task in Notion to add them when they're ready

css/card.css
```diff
@@ -90,6 +90,7 @@ tcg-card {
   margin: 0.5em 0;
   text-align: center;
   font-size: 1em;
+  height:20px;
```
miffens: why is a hardcoded height needed for the pull announcement?

index.html
```diff
@@ -86,19 +86,36 @@ <h3>Let's see what the Altventurers have made for you</h3>
       </section>
 
       <footer>
-        <a href="https://github.com/elysiumdelivery/altare-tcg"
-          ><svg
-            role="img"
-            viewBox="0 0 24 24"
-            xmlns="http://www.w3.org/2000/svg"
+        <div id="footer-nav">
+          <a id="footer-download-pdf" class="download-button" href="#">
```
miffens, line 90: Are the buttons on the fancy menu super big on purpose? If so, could the link text be aligned at the bottom for both instead of just one?

index.html
```diff
@@ -86,19 +86,36 @@ <h3>Let's see what the Altventurers have made for you</h3>
...
+              src="../icons/Icons_PDF.svg"
+            />
+            View Accessible Project Format</a
+          >
+        </div>
+        <p>Made with love by the Altventurers.</p>
+
+        <a href="https://github.com/elysiumdelivery/altare-tcg">
+          <svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" >
+              <title>GitHub</title>
```
miffens, line 111: could you change the link color to contrast better with the dark blue of the site (probably  white)?

pages/collection.html
kirinarimasu: Yep, that's correct.

css/card.css
kirinarimasu: skycompass commented that the height of the section kept jumping around. This specifically was to address the 'jumping' height after initial pull, as when the pull announcement is not populated it only has margin/padding and thereafter changes in height once text is pushed in.

index.html
kirinarimasu: No, something broke here - sorry. Nice catch. I will fix.

kirinarimasu: Updated on the homepage and made it fit in better with the existing splash page setup and animation.

index.html
kirinarimasu: Same issue as above - broken code. Updated now. :)

css/card.css
miffens: can you comment in code explaining that the height is so the section doesn't jump when text first appears?

kirinarimasu: Thanks, good idea!

miffens (approved):
Looks good, feel free to merge once the extra gap in Credits we're discussing on Discord is resolved

PR #40, opened 2023-01-25
General bug fixes & Nice to Haves

bonbombs:
This branch addresses the following fixes:

- Project Credit Order (Twitter Images section was sometimes loading before or after the credits sections, therefore messing up the css :nth-of-child() selectors, so I added ids to the credits sections and adjusted the selectors to point to those ids specifically) (merged by #39)
- Chibi Altare Disappears For a Moment
- Reanimate Card on Keyboard Focus


css/nav.css
```diff
@@ -57,7 +57,7 @@ nav a#nav-home #chibi-altare {
 
   transform-origin: 50% 100%;
 
-  transition: 0.5s opacity ease, 0.5s transform ease;
+  /* transition: 0.5s opacity ease, 0.5s transform ease; */
```
miffens: delete

js/cards.js
```diff
@@ -170,11 +179,10 @@ function setupHover() {
   // setup the initial event listeners for all cards
   let cardList = document.getElementsByClassName("card-image");
   for (let i = 0; i < cardList.length; i++) {
-    cardList[i].addEventListener("animationend", function () {
-      cardList[i].classList.remove("animated");
-      cardList[i].style.willChange = "auto";
-    });
+    cardList[i].addEventListener("animationend", removeAnimation);
+    cardList[i].parentElement.addEventListener("blur", removeAnimation);
```
miffens, line 188: I read the blur event to mean "cancel the animation when you keyboard focus away" but that doesn't seem to happen, so I'm not 100% sure what purpose this event listener is? :0

miffens (approved):

miffens:
oh and technically not in this PR anymore, but would you mind adding a comment in the code explaining the 250px thing that fixed the bottom gap on the credits page?

js/cards.js
bonbombs: oh oop. I forgot to check if the target was the button element or not for removeAnimation (like in addAnimation). Thank you for catching this!

PR #41, opened 2023-01-25
Dark Mode

bonbombs:
Currently just setting up colors & experimenting w/ them. Should be viewable if your browser is set to prefers dark mode

Still thinking about where to add the dark toggle though


miffens:
On a lot of light mode screens, the text is still white making it hard to see:
 - Card dialog window
 - Card balance disclaimer
 - Welcome message before the fancy menu

The light blue on button hover seems to no longer work in dark mode

miffens:
With little time left in the project I'd prefer to focus on the Dark Reader bugs and not offer an official dark mode toggle since it might risk regressing the colors of the overall site too much

Even the Dark Reader bugs are optional imo, it's a 3rd party extension that folks using it should be aware can mess with sites unintentionally

PR #42, opened 2023-01-25
Update card details dialog

miffens:
Show image description
Show attack cost/ damage if available
On top of o5v3's changes for a comprehensive preview. See https://github.com/elysiumdelivery/altare-tcg/commit/8172a93f6cad6f3311a03975f0d3c9dc96611427 for only my diff
I don't love how messy this is but I don't know a significantly better + quick way :c

### DIFF
--- a/js/dialog.js
+++ b/js/dialog.js
@@ -42,14 +42,14 @@ function attackTraitsHTML(cost, damage, type) {
     return "";
   }
   const htmlList = [];
+  if (type) {
+    htmlList.push(`Type: ${type}`);
+  }
   if (cost) {
     htmlList.push(`Cost: ${cost}`);
   }
   if (damage) {
     htmlList.push(`Damage: ${damage}`);
-  }
-  if (type) {
-    htmlList.push(`Type: ${type}`);
   }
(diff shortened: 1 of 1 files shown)


js/dialog.js
```diff
@@ -37,6 +37,23 @@ export async function setupDetailsDialog() {
...
+  }
+  if (cost) {
+    htmlList.push(`Cost: ${cost}`);
+  }
+  if (damage) {
+    htmlList.push(`Damage: ${damage}`);
+  }
+  return `<p>${htmlList.filter((item) => Boolean(item)).join(" | ")}</p>`;
+}
+
```
o5v3, line 56: There's a good way to compress this: Instead of receiving individual parameters, you can receive an object with the name of the variables as keys. That way you can use object traversal methods which can eliminate the repetition.

o5v3 (approved):
That's about the only change I see. Everything else is html templating so not much can be done about it. Feel free to merge this, I can do the change if you want.o7

js/dialog.js
miffens: hmm can you customize the label ('Type:') easily though? I'll go ahead and merge and leave this in your capable hands!

PR #45, opened 2023-01-26
Mobile QA

bonbombs:


css/card.css
```diff
@@ -570,7 +570,6 @@ section#gacha-controls fieldset:nth-child(2) > input:first-child {
   margin: 20px;
   overflow: hidden;
   z-index: 10;
-  touch-action: none;
```
bonbombs, line 573: @kirinarimasu I am not sure if removing this affects anything. This was causing the user to be unable to scroll on mobile. The card was eating the scroll event

css/nav.css
```diff
@@ -177,6 +177,8 @@ nav a#nav-credits:focus-visible:not(.active)::before {
   /* portrait / small width styling, show burger menu */
   #nav-hamburger {
     display: block;
+    /* ensure that this always stays up top */
+    z-index: 999;
```
miffens: will this show above the dialog then?

bonbombs: Ah shoot, you're right. I will need to adjust the dialog z-index also so it's greater than the hamburger button, which has to be greater than the max z-index value of the last flipped card in the gacha page.

css/card.css
kirinarimasu: I think this is okay to remove, thanks! It doesn't make sense that it's there anymore, nice catch!

js/cards.js
```diff
@@ -186,9 +186,9 @@ function setupHover() {
   for (let i = 0; i < cardList.length; i++) {
     cardList[i].addEventListener("animationend", removeAnimation);
     cardList[i].parentElement.addEventListener("blur", removeAnimation);
-    cardList[i].addEventListener("mouseover", addAnimation);
     cardList[i].parentElement.addEventListener("focus", addAnimation);
-    cardList[i].addEventListener("click", addAnimation);
+    cardList[i].parentElement.addEventListener("click", addAnimation);
+    cardList[i].parentElement.addEventListener("mouseover", addAnimation);
```
miffens: Not sure if this change is what caused it since it looks like there was only rearranging here, but in gacha, the cards below the topmost one are reanimating on hover

This doesn't seem to be happening on the main branch (at least on hover, I seem to be able to click into cards below causing it to reanimate)

bonbombs: I put the order back to where it was, but it's also now targeting the parent (button) element instead of the card-image div. Was trying to finagle which element to capture events so that a user could scroll on mobile while touching the cards.

miffens:
I'm seeing
```
Uncaught TypeError: Cannot read properties of undefined (reading 'parentElement')
    at removeUnclickableFromCards (cards.js:299:20)
    at HTMLInputElement.updateGachaView (cards.js:205:5)
```
when switching from card list to card grid view after flipping over a few cards

miffens (approved):

PR #46, opened 2023-01-26
Collections Page Dialog Updates

kirinarimasu:
General updates to make the card not clip the end of the dialog on shorter screens, and allow a larger card image on larger screens.


kirinarimasu:
Not sure if I merged correctly before - I can copy the code into a new branch if necessary...

kirinarimasu:
Too messy, so going to leave this.

PR #49, opened 2023-01-26
Do some quick fixes and support pretty urls

o5v3:
- Change the way to check the current page to support urls that don't end in ".html". It supports with and without the extension.
- Refactor the attackTraitsHTML function in dialog.js. It supports an object argument and works in the same way as the original but allows for any amount of traits to be passed to the function, including traits with html formatting or string interpolation.
- Allow the controls on the collections page to work even if something goes wrong in the setup process.


js/main.js
```diff
@@ -133,10 +133,10 @@ async function main() {
         }
         break;
 
-      case "/collection.html":
+      case "/collection":
+        setupCollectionControls();
```
miffens: do we need to change the other places with `*.html`? I think it doesn't break anything if we don't update all the places, but messages.js has some similar switch cases set up to this that would be good to make consistent

All the hardcoded urls in each page's html also still have an href of `*.html`

miffens (approved):

miffens (changes requested):
Hm hang on some traits data seems to be missing from some dialogs
001 Antare is missing the skill cost for "rise up, antnation"
002 Dear Astrobunny is missing the skill cost

Cards with hidden abilities (ie 006) are not showing "hidden ability" which should be on the Type attribute

js/main.js
o5v3: Thanks for the heads-up! Changed the routing in messages.js too.

If we turn on the Pretty URLs setting in Netlify, it searches for urls in the html files and automatically rewrites them without the html extension, so nothing has to be changed in that aspect. I want to wait until all branches get this routing change before activating that option, just to be safe.

o5v3:
> Hm hang on some traits data seems to be missing from some dialogs 001 Antare is missing the skill cost for "rise up, antnation" 002 Dear Astrobunny is missing the skill cost
> 
> Cards with hidden abilities (ie 006) are not showing "hidden ability" which should be on the Type attribute

Good catch. After looking through the code again I misunderstood some parts of the original function. Changed it and now it works the same way as the original. o7

miffens (approved):

miffens:
oh yeah, could you move the reset button code in this change as well?

Edit: nvm you did

PR #51, opened 2023-01-26
Collections Page Dialog Updates V2

kirinarimasu:
Copied dialog code updates from https://github.com/elysiumdelivery/altare-tcg/pull/46

### DIFF
--- a/css/card.css
+++ b/css/card.css
@@ -295,7 +295,7 @@ section label {
   animation: button-wiggle 2.32s ease 0s infinite;
 }
 .pile-display button.gacha-button {
-  margin-bottom:0.5em;
+  margin-bottom: 0.5em;
 }
 
 #gacha-controls button.gacha-button:hover,
@@ -311,18 +311,18 @@ section#gacha-controls label {
 section#gacha-controls input {
   margin-left: 15px;
 }
-section#gacha-controls fieldset:nth-child(2){
-  margin-left:20px;
-  border-left:1px solid var(--bg-color-lighten-50);
+section#gacha-controls fieldset:nth-child(2) {
+  margin-left: 20px;
+  border-left: 1px solid var(--bg-color-lighten-50);
 }
 section#gacha-controls fieldset:nth-child(2) > input:first-child {
-  margin-left:20px;
+  margin-left: 20px;
 }
 
 .grid-display .gacha-prompt {
   /* this is also part of the hack to ensure that the gacha section tag doesn't change
   as much when switching between display methods */
-  margin-top:1.5em;
+  margin-top: 1.5em;
 }
 .gacha-prompt {
   display: flex;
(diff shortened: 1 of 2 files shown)


css/card.css
```diff
@@ -389,6 +389,50 @@ section#gacha-controls fieldset:nth-child(2) > input:first-child {
   display: none;
 }
 
+@media screen and (min-width: 1600px) {
+  .details-dialog-card {
```
miffens: there are a lot of details-dialog CSS rules throughout this file and I wonder if they can be consolidated? Don't spend too much time though since we are short on time

css/card.css
```diff
@@ -389,6 +389,50 @@ section#gacha-controls fieldset:nth-child(2) > input:first-child {
...
+    overflow: auto;
+  }
+
+  .details-dialog-card {
+    /* make the card sit up the top instead */
+    align-self: start;
+  }
+
+  .details-dialog-text {
+    /* Override overflow: auto for small screens,
```
miffens: The independent scrolling of the text sections for large dialogs no longer seems to be working - the whole dialog scrolls now and the card no longer remains "locked" in view
If you can't figure it out in time, you can merge - I think this is preferred over a bit of text being unreadable

kirinarimasu: I can't seem to reproduce this sorry. When I try to test this the text is never long enough to be scrollable, but the card is too big; therefore the whole thing ends up scrolling. Maybe the card can be smaller after all? 700px feels a bit uncomfortable to me somehow.

(Edit: to be clear, the 700px is the card width. The height is something like 980px.)

miffens: don't quite follow your reply - maybe it was in response to the code I highlighted which may have nothing to do with the actual issue I was pointing out 😅 
This is the current dialog without the changes in this PR - no matter where the text is scrolling the card is always on screen
current dialog
In this PR, the card can scroll out of view
new dialog

kirinarimasu: Yeah I think it's to do with the card height, since you can see the top and bottom spacing around the card in the top screenshot are slightly different heights - I wanted it scroll when the above/below spacing became uneven. But I can see I probably need to be MUCH MORE specific in the set sizing, and it can be done better. I will revisit. :)
(Edit for missing word, clarity.)

css/card.css
```diff
@@ -389,25 +389,94 @@ section#gacha-controls fieldset:nth-child(2) > input:first-child {
...
+    background-size: 120% !important;
+    background-repeat: repeat;
   }
+}
 
+@media screen and (max-width: 1400px) {
+  #gacha-controls {
+    /* make the card smaller when the screen is resized */
+    --card-width: 300px;
+    --card-height: 419px;
```
miffens, line 469: (nit) is this 419 height 300px * 1.395? express it like calc to be consistent with the rest of the rules? else explain the slightly different math

miffens (approved):

css/card.css
kirinarimasu: The strange thing with the card sizing is that it's all based off the original 1488 × 2079 card exports (1.397 ratio between the two dimensions), but as you get smaller you start calculating fractions of pixels. If you get the dimensions from a card resize in Photoshop for example, you get 300x419 rather than 300x419.15. Will take a look at cleaning this up.

PR #55, opened 2023-01-27
Site Link Updates

kirinarimasu:
Linkage

### DIFF
--- a/index.html
+++ b/index.html
@@ -108,7 +108,7 @@
             />
             Download Full Project PDF</a
           >
-          <a id="footer-download-a11y" class="download-button" href="#">
+          <a id="footer-download-a11y" class="download-button" href="https://docs.google.com/document/d/1JNKdL4Js6JpJqumtm6uyyYqW_xKRv2OhT6fgoJWwK3g">
             <img
               class="icon icon-download"
               aria-hidden="true"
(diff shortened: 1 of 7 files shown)


index.html
```diff
@@ -108,7 +108,7 @@ <h3>Let's see what the Altventurers have made for you</h3>
             />
             Download Full Project PDF</a
           >
-          <a id="footer-download-a11y" class="download-button" href="#">
+          <a id="footer-download-a11y" class="download-button" href="https://docs.google.com/document/d/1JNKdL4Js6JpJqumtm6uyyYqW_xKRv2OhT6fgoJWwK3g">
```
miffens: assuming  this is still pending the pdf download link

could you also add target="_blank" to each of the links so they open in a new tab?

kirinarimasu: Gotcha, thanks!

miffens (approved):

js/messages.js
```diff
@@ -31,7 +31,9 @@ function renderMessages(message_data) {
     if (message["Twitter"]) {
       let userSocial = document.createElement("a");
       userSocial.href = `https://twitter.com/${message["Twitter"]}`;
+      userSocial.setAttribute('target', '_blank');
       userSocial.textContent = message["Twitter"];
+      userSocial.title = "Twitter page for " + message["Participant Name"] + " (" + message["Twitter"] + ")";
```
miffens, line 36: you could delete these - titles are actually not beneficial to accessibility since they're only activated with mouse and do nothing for keyboard/ screenreader

